PR #173, opened 2023-06-27
Add IC_BLevels parameter to CX and Varobs writers

ctgh:
Add the `IC_BLevels` parameter, which specifies the number of model boundary layer levels.

This is necessary for SHPSYN data, for which the wind speed is interpolated to 20m height in VAR. `IC_BLevels` must be present in the CX header for the interpolation to work correctly.

Add the same parameter to the VarObs writer for symmetry and to avoid unintended consequences.


ldhawknesssmith (approved):
Well done finding this. Looks good to me.

brettcandy (approved):
looks fine

ctgh:
Thanks for your reviews!

PR #174, opened 2023-06-29
Add CX and Varobs for Geos

PJLevensMO:
Added namelists to create Varobs and CX for the three Geo instruments (SEVIRI clear, ABI clear, and AHI clear) and the unit tests for all three


mikecooke77 (approved):
All looks good to me

brettcandy (approved):
this looks fine. I have checked the var obs namelist with what is expected for Var and it is consisntent.

smnewman (approved):

PR #180, opened 2023-09-05
update ci configuration to run on self-hosted runner

yaswant:
Update: CI build and test should now use Met Office self-hosted runner


.github/workflows/ci.yml
```diff
@@ -8,54 +8,56 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: jedi-self-hosted-rg
```
mikecooke77, line 11: This name seems a bit hard to interpret to me.  Why not:
jedi-self-hosted-runner

ci/build-and-test
```diff
@@ -0,0 +1,50 @@
...
+THIS="$(basename "$0")"
+NPROC=${NPROC:-$(nproc)}
+WORKD="$(mktemp -d "${THIS}-XXXXXX" -t)"
+
+trap finally ERR
+trap finally EXIT
+
+cd "${WORKD}"
+
+# - jcsda container requires manual sourcing of spack environment
```
mikecooke77: Do we need to keep these comment lines?

.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-name: CI
+name: UKMO_CI_Az
```
mikecooke77: Why not a little more explicit:
UKMO_CI_Azure

mikecooke77 (commented):
Looks good just a couple of comments

matthewrmshin (approved):
LGTM.

.github/workflows/ci.yml
yaswant: We have two action runner groups (rg): 
1. actions-rg (for MetOffice repos) 
2. actions-rg-ext (for externally hosted repos)

I've randomly chosen the names (jedi-self-hosted-rg, jedi-self-hosted-ext) over default action-runner during configuration time.  In order to change these, we need to go through the process again.. Hopefully we can sort this out in a future PR.

ci/build-and-test
yaswant: No. removed now.

.github/workflows/ci.yml
yaswant: Done.

mikecooke77 (approved):

PR #181, opened 2023-09-25
Map channels to Var Channels

orlewis:
This PR adds the functionality to map channels defined in JOPA to channels in Var for the varobs writer. 

As part of this change the offset channels option has been removed as this can nw be covered using the varChannels mapping option. 
use_actual_channels has been replaced by a compress_var_channels. This determines whether to compress channel numbers together i.e 3, 5, 100 in just three indices or if do not compress this would have an array of size 100. 

This will require changes to radiance yamls in Sith. 

Requires this change to be coordinated.
https://github.com/MetOffice/sith/pull/262


orlewis:
I think there is still a need to put in some failure captures, incompatible options etc. But thought should have some more eyes take a look at it first.

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -2279,13 +2299,30 @@ function opsinputs_fill_varnames_with_channels(VarName, Channels) result(VarName
...
+!if (Varname=="emissivity") then
+!  do ichan = 1, size(Channels)
+!    write(*,*) "here"
+!    write(*,*) VarName, Channels(ichan)
+!    write (VarNames_emis(ichan),'(A,"_",I0)') VarName, Channels(ichan)
+   ! write(*,'(A,"_",I0)') VarName, Channels(ichan)
+   ! write (VarNames_emis(ichan),'(A,"_",I0)') VarName, Channels(ichan)
+  !end do
+ ! VarNames = VarNames_emis
+!else
```
orlewis: Can't remember why I have commented this out but also there are no ctest failures?

src/opsinputs/VarObsWriter.cc
```diff
@@ -70,6 +71,12 @@ VarObsWriter::VarObsWriter(ioda::ObsSpace & obsdb, const Parameters_ & params,
   const int fallbackChannels = 0;
   // Avoid passing a null pointer to Fortran.
   const int *channelsData = channels.empty() ? &fallbackChannels : channels.data();
+
```
mikecooke77, line 74: Worth adding a comment here?

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1046,18 +1048,21 @@ subroutine opsinputs_fill_fillreal2d_norecords( &
...
+logical, optional, intent(in)                       :: compressVarChannels
+integer(integer64), optional, intent(in)            :: sizeVarObs
+integer(c_int), optional, intent(in)                :: varChannels(:)
+
 
 ! Local declarations:
 real(kind=c_double)                             :: VarValue(NumObs)
 real(kind=c_double)                             :: MissingDouble
-character(len=max_varname_with_channel_length)  :: JediVarNamesWithChannels(max(size(Channels), 1))
+character(len=200)  :: JediVarNamesWithChannels(max(size(Channels), 1))
```
mikecooke77: Why 200 rather than `max_varname_with_channel_length`?

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1046,18 +1048,21 @@ subroutine opsinputs_fill_fillreal2d_norecords( &
...
 integer                                         :: iChannel
 integer                                         :: offset
 integer                                         :: numchans
-logical                                         :: localUseActualChans
+integer                        :: offsetsize
```
mikecooke77: alignment of ::

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1046,18 +1048,21 @@ subroutine opsinputs_fill_fillreal2d_norecords( &
...
 
+!compressVarChannels=.TRUE.
```
mikecooke77: If this is a comment line is it still needed?

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1067,41 +1072,58 @@ subroutine opsinputs_fill_fillreal2d_norecords( &
...
   do iChannel = 1, size(JediVarNamesWithChannels)
-    ! Retrieve data from JEDI
-    call obsspace_get_db(ObsSpace, JediVarGroup, JediVarNamesWithChannels(iChannel), VarValue)
-
-    ! Fill the OPS data structures
-    if (localUseActualChans) then
-      where (VarValue /= MissingDouble)
-        Real2(:, Channels(iChannel)) = VarValue
-      end where
+      call obsspace_get_db(ObsSpace, JediVarGroup, JediVarNamesWithChannels(iChannel), VarValue)
```
mikecooke77: alignment with if statement below.

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1067,41 +1072,58 @@ subroutine opsinputs_fill_fillreal2d_norecords( &
...
+
+    if (present(varChannels)) then
```
mikecooke77, line 1110: I think a comment is needed to explain what is happening in this new section.

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1067,41 +1072,58 @@ subroutine opsinputs_fill_fillreal2d_norecords( &
...
+            end where
+          end if
+        end if
+      end if
     else
-     ! Fill the OPS data structures
-      where (VarValue /= MissingDouble)
-        Real2(:, iChannel+offset) = VarValue
-      end where
+      if ((present(sizeVarObs)) .and. (sizeVarObs > size(channels))) then
```
mikecooke77: These should be on separate clauses.  There is no guarantee in Fortran that the first clause is evaluated before the second.

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1067,41 +1072,58 @@ subroutine opsinputs_fill_fillreal2d_norecords( &
...
-    if (localUseActualChans) then
-      where (VarValue /= MissingDouble)
-        Real2(:, Channels(iChannel)) = VarValue
-      end where
+      call obsspace_get_db(ObsSpace, JediVarGroup, JediVarNamesWithChannels(iChannel), VarValue)
+
+    if (present(varChannels)) then
+      if (size(varChannels) > 0) then
+        if (iChannel <= size(varChannels)) then
+          if (compressVarChannels) then
```
mikecooke77: You should check in another if loop whether this is present or create a local variable with a default for use in the loop e.g.
``` Fortran
compressChannels = .false.
if (present(compressVarChannels)) then
  compressChannels = compressVarChannels
end if
```

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1067,41 +1072,58 @@ subroutine opsinputs_fill_fillreal2d_norecords( &
...
+          end if
+        end if
+      end if
     else
-     ! Fill the OPS data structures
-      where (VarValue /= MissingDouble)
-        Real2(:, iChannel+offset) = VarValue
-      end where
+      if ((present(sizeVarObs)) .and. (sizeVarObs > size(channels))) then
+        if (compressVarChannels) then
```
mikecooke77: probably safer with a local variable.

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1236,28 +1258,25 @@ subroutine opsinputs_fill_fillreal2d( &
...
-        JediVarName, JediVarGroup, OffsetChans, useActualChans)
-    else
-      call opsinputs_fill_fillreal2d_norecords( &
-        Hdr, OpsVarName, JediToOpsLayoutMapping % NumOpsObs, Real2, ObsSpace, Channels, &
-        JediVarName, JediVarGroup, OffsetChans)
-    end if
+  if (size(varChannels) > 0) then
+    call opsinputs_fill_fillreal2d_norecords( &
+      Hdr, OpsVarName, JediToOpsLayoutMapping % NumOpsObs, Real2, ObsSpace, Channels, &
+      JediVarName, JediVarGroup, compressVarChannels, sizeVarObs, varChannels)
```
mikecooke77: As `varChannels` , `sizeVarObs` and `varChannels` are all optional this code is currently not safe.  With optional there needs to be a present check else you can't guarantee memory integrity.

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1853,7 +1872,8 @@ subroutine opsinputs_fill_filltimeoffsets2d_norecords( &
 ! Local declarations:
 integer(c_int64_t)                              :: VarValue(NumObs)
 real(kind=c_double)                             :: MissingDouble
-character(len=max_varname_with_channel_length)  :: JediVarNamesWithChannels(max(size(Channels), 1))
+!character(len=max_varname_with_channel_length)  :: JediVarNamesWithChannels(max(size(Channels), 1))
+character(len=200)  :: JediVarNamesWithChannels(max(size(Channels), 1))
```
mikecooke77: Again why change to 200?

src/opsinputs/opsinputs_fill_mod.F90
mikecooke77: It would be good to understand this.

src/opsinputs/opsinputs_utils_mod.F90
```diff
@@ -40,6 +40,14 @@ module opsinputs_utils_mod
...
 end type opsinputs_channeloffset
 
+
+! Type to contain the channels offset when converting from JEDI to VAR.
+type, public :: opsinputs_varchannels
+  ! This is the size of the varobs array for output.  The default is zero and the size of the array will be used.
+  ! For atovs, jedi has 20 brightness temperatures but var expects 40.
+  ! Therefore for atovs brightness_tmperatuere => size_of_varobs_array = 40.
+  integer  :: size_of_varobs_array
+end type opsinputs_varchannels
```
mikecooke77: Given this only has one variable in this doesn't need to be a type anymore.

src/opsinputs/opsinputs_varobswriter_mod.F90
```diff
@@ -186,15 +186,17 @@ module opsinputs_varobswriter_mod
   integer(integer64) :: VarobsLength
 
   integer(c_int), allocatable :: channels(:)
+ ! integer(c_int), allocatable :: jopaChannels(:)
```
mikecooke77: remove?

src/opsinputs/opsinputs_varobswriter_mod.F90
```diff
@@ -186,15 +186,17 @@ module opsinputs_varobswriter_mod
...
+  integer(c_int), allocatable :: varChannels(:)
+  integer(integer64) :: size_of_varobs_array
+  integer(integer64) :: NumVarChannels
+
+  logical ::compressVarChannels
```
mikecooke77: sapce after ::

src/opsinputs/opsinputs_varobswriter_mod.F90
```diff
@@ -1166,9 +1161,11 @@ subroutine opsinputs_varobswriter_populateobservations( &
   end select
 
   if (FillChanNum .or. FillNumChans) then
+
```
mikecooke77: remove blank line

src/opsinputs/opsinputs_varobswriter_mod.F90
```diff
@@ -1213,54 +1210,104 @@ end subroutine opsinputs_varobswriter_fillreportflags
...
 type(OB_type), intent(inout)   :: Ob
 type(c_ptr), value, intent(in) :: ObsSpace
-integer(c_int), intent(in)     :: Channels(:)
+integer(c_int), intent(in)     :: channels(:)
+integer(c_int), intent(in)     :: varChannels(:)
 type(c_ptr), value, intent(in) :: Flags
 logical, intent(in)            :: FillChanNum, FillNumChans
-integer, intent(in)            :: OffsetChans
-logical, optional, intent(in)  :: UseActualChans
+logical                     :: compressVarChannels
```
mikecooke77: alignment, optional?

src/opsinputs/opsinputs_varobswriter_mod.F90
```diff
@@ -1213,54 +1210,104 @@ end subroutine opsinputs_varobswriter_fillreportflags
...
 ! Local declarations:
 integer(integer64)             :: NumChannels
-integer(integer64)             :: ChannelIndices(Ob % Header % NumObsLocal, size(Channels))
+integer(integer64), allocatable            :: ChannelIndicesVar(:,:)
+integer(integer64)             :: ChannelIndices(Ob % Header % NumObsLocal, size(channels))
 integer(integer64)             :: ChannelCounts(Ob % Header % NumObsLocal)
 integer                        :: iChannel
 integer                        :: iObs
-logical                        :: localUseActualChans
+real(kind=c_double)                             :: MissingDouble
```
mikecooke77: alignment

src/opsinputs/opsinputs_varobswriter_mod.F90
```diff
@@ -1213,54 +1210,104 @@ end subroutine opsinputs_varobswriter_fillreportflags
...
+MissingDouble = missing_value(0.0_c_double)
+
+NumChannels = size(channels)
 if (NumChannels == 0) return
 
-! Setup for useActualChans for array indices
-localUseActualchans = .false.
-if (Present(UseActualChans)) then
-  localUseActualChans = UseActualChans
+if (increaseChanArray) then
```
mikecooke77: again this routine is not safe with its use of optional arguments without checking they are present.  I would suggest creating local variables from all optional.

test/testinput/010_VarField_britemp.yaml
```diff
@@ -28,38 +28,37 @@ observations:
...
-        Callsign: ["station_1", "station_1",
-                   "station_2", "station_2",
-                   "station_3", "station_3",
-                   "station_4", "station_4"]
+        field:    ["10", "10","54","55","55",
+                   "10", "10","54","55","55",
+                   "10", "10","54","55","55",
+                   "10", "10","54","55","55"]
+        level:    ["1", "2","1","1","2",
+                   "1", "2","1","1","2",
```
mikecooke77: why have these values changed?

test/testinput/010_VarField_britemp.yaml
```diff
@@ -90,38 +89,262 @@ observations:
...
+  - obs space:
+      name: AMSUB
+      obsdatain:
+        engine:
+          type: H5File
+          obsfile: /data/users/olewis/JEDI_branches/test_geo_opsinputs/build-mo-bundle-jopa_env-RelWithDebInfo/opsinputs/test/Data/010_VarField_britemp.nc4 #Data/010_VarField_britemp.nc4
+      obsdataout:
+        engine:
+          type: H5File
+          obsfile: /home/h01/olewis/map_chans.nc4
```
mikecooke77: own paths

test/testinput/010_VarField_britemp.yaml
```diff
@@ -90,38 +89,262 @@ observations:
...
+
+#
+# Now we test effect of var channels
+#
+  - obs space:
+      name: AMSUB
+      obsdatain:
+        engine:
+          type: H5File
+          obsfile: /data/users/olewis/JEDI_branches/test_geo_opsinputs/build-mo-bundle-jopa_env-RelWithDebInfo/opsinputs/test/Data/010_VarField_britemp.nc4 #Data/010_VarField_britemp.nc4
```
mikecooke77: own paths

mikecooke77 (changes requested):
I have done a first look through.  I think the general code change is OK.  I have suggested some changes and will re-review after changes.

DJDavies2:
This isn't compiling for me:

/home/h01/david.davies/cylc-run/opsinputs-181/share/mo-bundle/opsinputs/src/opsinputs/opsinputs_fill_mod.F90:1070:4:

 1070 | if present(compressVarChannels) then
      |    1
Error: Missing '(' in IF-expression at (1)

src/opsinputs/opsinputs_fill_mod.F90
orlewis: fixed

src/opsinputs/opsinputs_fill_mod.F90
orlewis: Removed

src/opsinputs/opsinputs_fill_mod.F90
orlewis: Have now changed

src/opsinputs/opsinputs_fill_mod.F90
orlewis: done

src/opsinputs/opsinputs_fill_mod.F90
orlewis: done

src/opsinputs/opsinputs_fill_mod.F90
orlewis: fixed

src/opsinputs/opsinputs_fill_mod.F90
orlewis: fixed

src/opsinputs/opsinputs_utils_mod.F90
orlewis: removed

src/opsinputs/opsinputs_varobswriter_mod.F90
orlewis: done

src/opsinputs/VarObsWriterParameters.h
```diff
@@ -79,21 +79,20 @@ class VarObsWriterParameters : public oops::ObsFilterParametersBase {
...
-  /// [NaN,NaN,NaN,Nan,5,6,7,Nan,9,10,11,NaN].
-  oops::Parameter<bool> use_actual_channels{"use_actual_channels", false, this};
+  /// List of channels which are expected for Var.
+  oops::Parameter<std::string> varChannels{"varChannels", "", this};
+
+  /// Compress the var channels
+  oops::Parameter<bool> compressVarChannels{"compress_var_channels", true, this};
+
+  /// Increase the channel array size to the same size as the varobs array
+  oops::Parameter<bool> increaseChanArray{"increase_chan_array", false, this};
```
mikecooke77: Could more detailed descriptions be added please.

orlewis:
@ctgh and @james-cotton, This is the PR for adding the functionality needed for the geostationary radiances. The ctests have been passing but when I ran with sith kgo testing I got varobs failures with Satwind and Sonde http://fcm1/cylc-review/view/olewis?&suite=sith_map_chans_kgo&no_fuzzy_time=0&path=log/job/20210701T1200Z/glu_jopa_compare_output/07/job.out
The change is mainly focused on how channels are manipulated and I had though that Satwind and Sonde are insulated from these particular changes but maybe I have missed something.

ctgh:
@orlewis  I believe James' recent optimisation PR in sith changed some of the KGOs due to using a different number of PEs. Also, this branch is out of date with develop. Finally I find that running `rose suite-clean` prior to running the KGO test can prevent some false positives from appearing.

src/opsinputs/opsinputs_varobswriter_mod.F90
```diff
@@ -235,6 +236,13 @@ function opsinputs_varobswriter_create(self, f_conf, comm_is_valid, comm, channe
 allocate(self % channels(size(channels)))
 self % channels(:) = channels(:)
 
+call f_conf % get_or_die("NumVarChannels", self % NumVarChannels)
+allocate(self % varChannels(self % NumVarChannels))
```
mikecooke77, line 241: If these are allocated here won't it mean that localvarChannels are always allocated even if size zero and therefore the if allocated in the fill routine will always be use this.

src/opsinputs/opsinputs_varobswriter_mod.F90
```diff
@@ -1213,54 +1208,122 @@ end subroutine opsinputs_varobswriter_fillreportflags
...
+        if (compressChannels) then
+          offsetsize = abs(varChannels(1) - channels(1))
+          Ob % ChanNum = ChannelIndices
+          where (Ob % ChanNum > 0)
+            Ob % ChanNum = Ob % ChanNum + int(offsetsize, kind=integer64)
+          end where
+        else
+          do iChannel=1,  NumChannels
+            if (iChannel > size(varChannels)) then
+              ChannelIndicesVar(:, iChannel) = -32768
```
mikecooke77: isn't there a missing ops int value somewhere as a parameter?

src/opsinputs/opsinputs_varobswriter_mod.F90
```diff
@@ -1213,54 +1208,122 @@ end subroutine opsinputs_varobswriter_fillreportflags
...
-      ChannelIndices(:,iChannel) = Channels(iChannel)
-    end do
-    Ob % ChanNum = ChannelIndices
+  if ((varObsSize_loc /= 0)) then
+    if (varObsSize_loc > size(channels)) then
+      if (size(varChannels) > 0 .and. (size(channels) == size(varChannels))) then
+        if (compressChannels) then
+          offsetsize = abs(varChannels(1) - channels(1))
+          Ob % ChanNum = ChannelIndices
+          where (Ob % ChanNum > 0)
```
mikecooke77, line 1295: Won't this where clause always be true?

mikecooke77 (commented):
A last few comments / questions

src/opsinputs/opsinputs_varobswriter_mod.F90
orlewis: This was taken from what was in this routine previously. The where is checking for missing values to only apply where there is not a missing value.

src/opsinputs/opsinputs_varobswriter_mod.F90
orlewis: RMDI looks like the one to use.

ctgh: I think IMDI? (Ops_Constants/GenMod_MiscUMScienceConstants.f90)

src/opsinputs/opsinputs_varobswriter_mod.F90
orlewis: Yes the local will be allocated but then there are checks for if the size of varChannels > 0. 

I think I might have started doing the varChannels as a must include option and have later switched it to an optional. 

It might be an idea to simplify the logic that we use varChannels as the writout option, if it has size zero is reallocated to the channels and filled with the jopa channel values. Then we could remove one of the if blocks which is checking if varChannels is greater then size zero.

orlewis:
KGO is now passing (aprat from GPSRO which I think is failing the nightly tests aswell) http://fcm1/cylc-review/view/olewis?&suite=sith_kgo_map_chans231020&no_fuzzy_time=0&path=log/job/20210701T1200Z/glu_jopa_compare_output/03/job.out

orlewis:
The latest KGO output is here http://fcm1/cylc-review/view/olewis?&suite=sith_map_chans_kgo_231023&no_fuzzy_time=0&path=log/job/20210701T1200Z/glu_jopa_compare_output/09/job.out all passing apart from GPSRO which is currently failing the main nightly KGO. 
Using the current nightly GPSRO varobs and cx output all comparisons match as seen in http://fcm1/cylc-review/view/olewis?&suite=sith_map_chans_kgo_231023&no_fuzzy_time=0&path=log/job/20210701T1200Z/glu_jopa_compare_output/11/job.out

src/opsinputs/VarObsWriterParameters.h
```diff
@@ -79,21 +79,25 @@ class VarObsWriterParameters : public oops::ObsFilterParametersBase {
...
-  oops::Parameter<bool> use_actual_channels{"use_actual_channels", false, this};
+  /// List of channels which are expected for Var. This enables mapping of JOPA
+  /// channel numbers to VAR numbers. Default is empty and the JOPA channels will be used.
+  oops::Parameter<std::string> varChannels{"varChannels", "", this};
+
+  /// If set to false (default true) this matches the channel number with the array index.
+  /// This is useful when there are skipped channels. e.g. channels 5,6,7,9,10,11 are
+  /// required from a possible 12 channels. This would fill a size 12 array with the array
+  /// indexes matching the channel number [NaN,NaN,NaN,Nan,5,6,7,Nan,9,10,11,NaN].
+  /// Compressed example could be [2,8,50,100], no missin values between channel numbers.
```
james-cotton: Typo on "missing"

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -2379,13 +2422,25 @@ function opsinputs_fill_varnames_with_channels(VarName, Channels) result(VarName
...
 integer                                        :: ichan
 
-if (size(Channels) == 0) then
-  VarNames(1) = VarName
-else
+character(len=max_varname_with_channel_length) :: VarNames_emis(max(size(Channels), 1))
+
+
+
+if (Varname=="emissivity") then
```
james-cotton: Can you explain why these changes are needed for emissivity? There are also extra blank lines above this section

src/opsinputs/opsinputs_utils_mod.F90
```diff
@@ -40,6 +40,8 @@ module opsinputs_utils_mod
   integer  :: size_of_varobs_array
 end type opsinputs_channeloffset
 
+
```
james-cotton: Extra whitespace

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1057,52 +1060,76 @@ subroutine opsinputs_fill_fillreal2d_norecords( &
...
-  offset = OffsetChans % channel_offset
-  if (OffsetChans % size_of_varobs_array > 0) &
-    numchans = OffsetChans % size_of_varobs_array
-end if
 
-!Setup for channels needing to match array index
-localUseActualChans = .false.
-if (present(useActualChans)) then
-    localUseActualChans = useActualChans
+if (present(sizeVarObs)) then
```
james-cotton: Is it worth adding a comment here explaining what `sizeVarobs` is? I can see it is used to set the number of channels, but its not clear to me where this has come from (sizeVarobs to me feels like it is the number of varobs, which it isn't)

test/testinput/010_VarField_britemp.yaml
```diff
@@ -125,3 +124,214 @@ observations:
...
+    # values are encoded correctly in the VarObsFile.
+    - filter: Reset Flags to Pass
+      flags_to_reset: [10, 15]  # missing, Hfailed
+    - filter: VarObs Writer
+      variables_for_quality_control:
+        - name: brightnessTemperature
+          channels: 1,3
+      namelist_directory: testinput/VarObsWriterNamelists_010_VarField_britemp
+      general_mode: debug
+      size_of_varobs_array: 3
```
james-cotton, line 159: I was expecting this to be 2, since we are compressing to levels ["2","4"]? Or am I misunderstanding what compressing channels does?

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1057,52 +1060,76 @@ subroutine opsinputs_fill_fillreal2d_norecords( &
...
+    ! if VAR channels have been assigned then jopa channels will be mapped to these var channels
+    ! Set up the size of the array, if channels are being pushed together an offset between
+    ! the var and jopa channel numbers is added onto the size of the array.
+    ! If not compressed the positions in the array are based on the actual channel number.
+
+    if (present(varChannels)) then
+      if (size(varChannels) > 0) then
+        if (iChannel <= size(varChannels)) then
+          if (compressChannels) then
+            offsetsize = abs(varChannels(1) - channels(1))
```
james-cotton, line 1114: I am not really understanding what is happening in this section. 
Also, this assumes a constant offset - is this true? What if we have 
```
channels: 1, 3
varChannels: 2, 5
```

james-cotton (commented):
Hi Owen, some comments and questions

src/opsinputs/opsinputs_fill_mod.F90
orlewis: This is setting the index positioning. The constant offset is replicating how the previous offset_channels worked, this included a number of empty spaces at the start the output array. So in the example you have provided it would set an offset of 1 so there would be an empty space at the start of the array and then channel2 and 5 in the subsequent two spaces.

test/testinput/010_VarField_britemp.yaml
orlewis: This is testing the case where you are compressing the channels to be next to each other so you would have 2,4 but where the size of the varobs array is 3 to have an empty slot.

src/opsinputs/opsinputs_fill_mod.F90
orlewis: It is the size of varobs array which can be set as an input parameter. I might have shorthanded too much and added a capitalised the o. I'll rename to sizeOfVarobsArray and add a comment.

test/testinput/010_VarField_britemp.yaml
james-cotton: Why is there an empty slot? I assumed compressing removed the empty slots

orlewis: An empty slot ahead in this case, or you might need empty slots behind the values for example airs has dummy channel spaces. 
In this case it is applying the offsetchans effect, offsetting the channels by 1 which leaves the empty space at the start.

src/opsinputs/opsinputs_fill_mod.F90
orlewis: It was required at one point to get the ctests to pass but I think some other changes have made it redundant. I have just tested removing this section and all the ctests continued to pass.

orlewis: https://github.com/MetOffice/opsinputs/commit/61c2116df1d68183cbeba5b23d6fa44e304c8706 at this point I re included it. I'm trying to figure out if it is required.

orlewis: Run a new set of kgo's with this emissivity part removed and got the same result as before. So have now removed. 
http://fcm1/cylc-review/view/olewis?&suite=sith_map_chans_kgo_231031&no_fuzzy_time=0&path=log/job/20210701T1200Z/glu_jopa_compare_output/02/job.out

james-cotton: Thanks for checking this!

src/opsinputs/opsinputs_fill_mod.F90
james-cotton: Thanks for renaming and clarifying

test/testinput/010_VarField_britemp.yaml
```diff
@@ -125,3 +124,214 @@ observations:
...
+      varChannels: 2, 4 #[5,7]
+    - filter: VarObs Checker
+      expected_main_table_columns:
+        # In the arrays below, rows denote locations and columns levels.
+        field:    ["10", "10", "10",
+                   "10", "10", "10",
+                   "10", "10", "10",
+                   "10", "10", "10"]
+        level:    ["1","2","3",
```
james-cotton, line 168: How would we know from the varobs that column 3 here is var channel 4?

test/testinput/010_VarField_britemp.yaml
```diff
@@ -125,3 +124,214 @@ observations:
...
+    - filter: VarObs Writer
+      variables_for_quality_control:
+        - name: brightnessTemperature
+          channels: 1,3
+      namelist_directory: testinput/VarObsWriterNamelists_010_VarField_britemp
+      general_mode: debug
+      compress_var_channels: false
+      size_of_varobs_array: 4
+      varChannels: 2, 4 #[5,7]
+      increase_chan_array: true
```
james-cotton, line 303: This option doesn't seem to have an impact (output is the same as the previous test). Does this option need a different test?

james-cotton (commented):
Hi Owen, coming back to this

test/testinput/010_VarField_britemp.yaml
orlewis: From line 1287 of opsinputs_varobswriter_mod.f90.
The routine follows how the offset channels functionality would have worked. So previously Ob % ChanNum = ChannelIndices and Ob %  ChanNum = Ob % ChanNum + offset. 
The offset is now calculated by the difference between jopa channels and var channels. This is a limitation on the channel mapping though.

james-cotton: OK so VAR is just expecting to receive a certain offset channel numbering?

orlewis: Yes, the ChanNum originally came from the ChannelIndices and the offset increased the cahnnel number. I've been thinking I might be able to remove the limitation by setting up an offset array and then any channel number could be mapped to.

james-cotton (approved):
Hi Owen, I am generally happy that the current behaviour is replicated as expected, but as noted there is room to improve the way this works in future.

ctgh (commented):
The routines `opsinputs_fill_fillreal2d_norecords` and `opsinputs_varobswriter_fillchannumandnumchans` now both contain long nested `if` blocks. What do you think about creating two helper functions to move some of that functionality outside of those functions? That might make the logic of the code a bit easier to follow and maintain. Just an idea - no problem if you think it won't be useful.

orlewis:
> The routines `opsinputs_fill_fillreal2d_norecords` and `opsinputs_varobswriter_fillchannumandnumchans` now both contain long nested `if` blocks. What do you think about creating two helper functions to move some of that functionality outside of those functions? That might make the logic of the code a bit easier to follow and maintain. Just an idea - no problem if you think it won't be useful.

I think this would be useful but also at this point I think we should get this code in and ready to use for the Geostationary radiances so they can be finished for the global.

ctgh (approved):
Sure, fine for me.

mikecooke77:
@orlewis there is some pressure to get the geos done before xmas.  Is this ready to go.  Has the latest code been re-run with the sith-kgo.  If not could this be set to run asap?

orlewis:
There is KGO from 31/10/23 but there has been one change since then which was adding IMDI. http://fcm1/cylc-review/taskjobs?per_page=15&user=olewis&no_fuzzy_time=0&suite=sith_map_chans_kgo_231031&task_status=runahead&task_status=waiting&task_status=held&task_status=queued&task_status=expired&task_status=ready&task_status=submit-failed&task_status=submit-retrying&task_status=submitted&task_status=retrying&task_status=running&task_status=failed&task_status=succeeded&page=2 there are failures related to GPSRO KGO but these were consistent with the nightly tests at the time. 

Because it has been a month I''l rebuild and run them again today.

chawnharlow:
Hi guys, I just need to let you know that David Simonin has requested that the GEOs get treated as priority which means that this PR should be seen as high priority. I've not been following the action here, but if you can wrap it up that would be greatly appreciated. Thanks.

ctgh:
@chawnharlow We have two approvals so this can be merged if the KGO tests are passing. I can do the merge this afternoon if everything is OK.

orlewis:
http://fcm1/cylc-review/view/olewis?&suite=sith_kgo_mapchans_231204&no_fuzzy_time=0&path=log/job/20210701T1200Z/glu_jopa_process_background_ssmis/01/job.out Getting this error across several obs on the JOPA run. Haven't been able to find where it is coming from this morning.

ctgh:
`varno number with the same dimension` was recently added to ioda. Therefore I suspect you need to update all of the repositories in your local mo-bundle, rebuild, and rerun the KGO tests.

orlewis:
KGO results are in http://fcm1/cylc-review/view/olewis?&suite=sith_kgo_mapchans_231204&no_fuzzy_time=0&path=log/job/20210701T1200Z/glu_jopa_compare_output/02/job.out 

All passed.

ctgh:
Good news. @mikecooke77 I have added the 'ready to merge' label if you would like to take one more look.

mikecooke77 (approved):

PR #183, opened 2023-09-28
fix ci workflow bug

yaswant:
Appears that I have broken the CI workflow while switching to use Met Office self-hosted runner.
Now I've added workflow to check against both JCSDA and MetOffice environment


matthewrmshin (approved):
Looks good. CI output looks sensible.

james-cotton (approved):

PR #185, opened 2023-10-09
Write H(x) into CX for sonde data in JADA

ctgh:
Radiosonde data are processed as follows:
* JOPA + VAR: ProfileAverage operator, write GeoVaLs into CX for assimilation.
* JOPA + JADA: VertInterp operator, send H(x) file to JADA.

In the JOPA + VAR configuration, the GeoVaLs written to CX correspond to the model column associated with the lowest-altitude observation in a radiosonde profile. This matches closely what is done in OPS + VAR. Due to the reduced resolution of VAR, horizontal drift is discarded, which is why using a single model column works. Furthermore VAR assigns the CX to H(x) without performing any vertical interpolation.

In order to compare JADA and VAR it is necessary to run JOPA in an identical configuration and then send the same information to the assimilation. At present, the observation operator and the information sent are both different, making a precise comparison difficult. The observation operator must be changed to VertInterp when using JADA due to an unfixable issue in the TL/AD code for the ProfileAverage operator. However, when VertInterp is used, the values of H(x) sent to JADA are different.

This PR adds another option to opsinputs (`write_hofx_into_cx`) which writes H(x) values into CX rather than GeoVaLs. That makes it possible to run the VertInterp operator in JOPA and send a consistent H(x) to JADA. In other words this can be thought of as a 'JADA compatibility mode'. By default `write_hofx_into_cx` is false so it will not alter the behaviour of sith, etc.

The PR actually restores code that was removed in #110. At that point this functionality was not needed. However @mikecooke77 asked the prescient question "Is this work going to be revisited before JADA?". The answer is yes!

### DIFF
--- a/src/opsinputs/opsinputs_cxwriter_mod.F90
+++ b/src/opsinputs/opsinputs_cxwriter_mod.F90
@@ -27,7 +27,7 @@ use opsinputs_cxfields_mod
 use opsinputs_fill_mod, only:                 &
     opsinputs_fill_fillreal,                  &
     opsinputs_fill_fillrealfromgeoval,        &
-    opsinputs_fill_fillreal2dfromgeoval
+    opsinputs_fill_fillreal2dfromgeovalorhofx
 use opsinputs_mpl_mod, only: opsinputs_mpl_allgather_integer
 use opsinputs_obsspace_mod, only: opsinputs_obsspace_get_db_datetime_offset_in_seconds
 use opsinputs_utils_mod, only: &
@@ -137,6 +137,7 @@ private
   logical                                :: RejectObsWithAnyVariableFailingQC
   logical                                :: RejectObsWithAllVariablesFailingQC
   logical                                :: GeoVaLsAreTopToBottom
+  logical                                :: WriteHofXIntoCX
 
   character(len=100)                     :: latitudeName
   character(len=100)                     :: longitudeName
@@ -256,6 +257,9 @@ call f_conf % get_or_die("reject_obs_with_all_variables_failing_qc", &
 
 call f_conf % get_or_die("geovals_are_top_to_bottom", &
                          self % GeoVaLsAreTopToBottom)
+
+call f_conf % get_or_die("write_hofx_into_cx", &
+                         self % WriteHofXIntoCX)
 
 call f_conf % get_or_die ("latitude_name", StringValue)
 self % latitudeName = StringValue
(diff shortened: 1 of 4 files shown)


src/opsinputs/opsinputs_cxwriter_mod.F90
```diff
@@ -1087,174 +1091,179 @@ subroutine opsinputs_cxwriter_populatecx(self, ReportFlags, Cx)
 
     ! Upper-air variables
     case (StashItem_theta) ! IndexCxtheta
-      call opsinputs_fill_fillreal2dfromgeoval( &
+      call opsinputs_fill_fillreal2dfromgeovalorhofx( &
         Cx % Header % theta, "theta", Cx % theta, &
         self % GeoVals, self % GeoVaLsAreTopToBottom, opsinputs_cxfields_theta, &
-        self % JediToOpsLayoutMapping, self % hofx, self % varnames)
+        self % JediToOpsLayoutMapping, self % hofx, self % varnames, &
+        "potentialTemperature", self % WriteHofXIntoCX)
```
ctgh, line 1098: Note the extra options provided to theta, T, u, v and RH. These options are not needed otherwise.

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1434,24 +1436,122 @@ subroutine opsinputs_fill_fillreal2dfromgeoval( &
...
+!>   Name of the OB_type field to which \p Real2 corresponds.
+!> \param[inout] Real2
+!>   Pointer to the array to be populated.
+!> \param[in] JediToOpsLayoutMapping
+!>   Mapping between indices of observations in the JEDI and OPS data structures.
+!> \param[in] hofx
+!>   The HofX vector to use.
+!> \param[in] GeoVals
+!>   A container holding the specified GeoVaL.
+subroutine opsinputs_fill_fillreal2dfromhofx( &
```
ctgh, line 1503: This subroutine was removed in #110 and has been restored.

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1420,8 +1421,9 @@ end subroutine opsinputs_fill_fillreal2dfromgeovalforsinglelevelobs
 !>   HofX data structure.
 !> \param[in] varnames
 !>   List of simulated variables.
-subroutine opsinputs_fill_fillreal2dfromgeoval( &
-  Hdr, OpsVarName, Real2, GeoVals, GeoVaLsAreTopToBottom, JediVarName, JediToOpsLayoutMapping, hofx, varnames)
+subroutine opsinputs_fill_fillreal2dfromgeovalorhofx( &
```
ctgh, line 1424: This subroutine was removed in #110 and has been restored.

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1434,24 +1436,122 @@ subroutine opsinputs_fill_fillreal2dfromgeoval( &
...
 ! Body:
 
 if (JediToOpsLayoutMapping % ConvertRecordsToMultiLevelObs) then
-   call opsinputs_fill_fillreal2dfromgeovalformultilevelobs( &
-        Hdr, OpsVarName, JediToOpsLayoutMapping % NumOpsObs, Real2, &
-        GeoVals, GeoVaLsAreTopToBottom, JediVarName, JediToOpsLayoutMapping)
+   ! Index of variable in hofx array.
+   hofxIndex = 0
+   do i = 1, size(hofx, 1)
+      if (present(JediObsName)) then
```
ctgh, line 1458: The use of `JediObsName` reproduces the name mapping between GeoVaL and observation names (e.g. relative_humidity -> relativeHumidity). This wasn't needed in the code removed by #110.

james-cotton (approved):
Hi @ctgh, I have compared these changes with what was removed in #110 and it all appears consistent, plus the addition of the new parameter option. Looks fine to be merged.

mikecooke77 (approved):
The change looks good to me.  I can't fault the code.  If you are ready for this to be merged then please change the flag.

ctgh:
Thanks for your reviews! I have updated the label.

PR #192, opened 2023-12-19
Add a smal change to bugfix map chans for seviri to work correctly.

orlewis:
When final testing was carried out on Seviri the VAR statistics did not look correct. This problem was only happening for channel 6. 

The bug comes about from seviri having 6 channels read in but only 5 output to the varobs. This meant that a loop which cycled on the number of channels went round 6 times. A variable called arrayindex was used to assign where in an array to palce a value, this is initially set as the loop number. For seviri the logic follows a route where the arrayindex was updated to be the actual channel number, however on the final loop 6 the if logic prevents the arrayindex being reallocated and so channel 6 gets filled with the final channel 11 observations. 

This is fixed by setting an exit on the loop if the condition is not met.

### DIFF
--- a/src/opsinputs/opsinputs_fill_mod.F90
+++ b/src/opsinputs/opsinputs_fill_mod.F90
@@ -1116,6 +1116,8 @@ if (obsspace_has(ObsSpace, JediVarGroup, JediVarNamesWithChannels(1))) then
           else
             arrayindex = varChannels(iChannel)
           end if
+	else
+	  exit
         end if
       else 
         if (.not. compressChannels) then


ctgh (approved):
Thanks, looks like a straightforward change.

PJLevensMO (approved):
Thanks for adding this fix. I have tested this change for SEVIRI, and it fixes the problem with channel 6. I have also checked the effect on the other Geo instruments (ABI and AHI) and they remain the same, as expected.

orlewis:
KGO have been run successfully http://fcm1/cylc-review/taskjobs/olewis/?suite=sith_kgo_231219

Geos can also be seen here:

SEVIRI: https://www-nwp/~plevens/ngops_boitenoire/sith_SEVIRI_test/seviriclr/glu_20210701T1200Z/
ABI: https://www-nwp/~plevens/ngops_boitenoire/sith_ABI_test/abiclr/glu_20210701T1200Z/
AHI: https://www-nwp/~plevens/ngops_boitenoire/sith_AHI_test/ahiclr/glu_20210701T1200Z/

PR #195, opened 2024-02-09
partial restriction on ci runner

yaswant:
Since opsinputs is now public, I am removing CI run on our self-hosted runner against pull_requests.  Instead, I've added a workflow_dispatch action which can be triggered manually by authors with write privilege.

The CI runs against PR will continue on GitHub runners using JCSDA's gnu and clang environment, as before.


ctgh:
@yaswant Can you explain what this is doing, please? I am not familiar with the syntax.

ctgh (approved):

mikecooke77 (approved):

Issue #155, opened 2023-02-16
Use explicit CX namelist for global surface observations


ldhawknesssmith:
[opsinputs/pull/152](https://github.com/MetOffice/opsinputs/pull/152) introduces the CX namelist for global surface observations. @ctgh suggested that it would be clearer to explicitly list the CX fields requested rather than using the add/remove option.

This should be a straightforward change to address with no change to the result.

ctgh:
Hi @ldhawknesssmith, @JoanneAWaller, @ReubenHill,

I think it might be good to modify the namelist as mentioned above. It will make it much easier to see what CX columns are actually used in the surface processing. For example, it will be clear that the surface columns differ from the screen ones.

Maybe we could discuss that in an upcoming meeting?

PR #199, opened 2024-04-16
Add cx namelist for UKV surface

ReubenHill:
This adds a cx namelist for UKV surface processing. This is almost identical  to the global cx namelist (#152), except that the cx fields are explicitly given as suggested in #155 . 

The fields were chosen by inspecting the ukv surface OPS `.varobs` files to see which fields are output. Since the present status of the UKV surface processing uses the same fields as the global processing (visibility has yet to be ported), those which are removed in #152 are removed here. I noted that not all of them have been ported to JOPA, but found that the VAR task in sith for both JOPA and OPS gave equivalent outputs when buddy check and thinning are switched off (i.e. within double vs single precision floating point arithmetic error.)

With thinning and buddy check turned back on, one yields the following analysis increment difference with [this version of sith](https://github.com/MetOffice/sith/tree/0067167a9e675d249b0c18eda5f8fe1546b1e8be), [this JJAUX](https://github.com/MetOffice/jjaux/tree/df2217dc66608a3cfd738fdce737ad6fd2564848) and [this jjdocs](https://github.com/MetOffice/jjdocs/tree/1f3a8d9b460459c4f8bf9e85f4c8de7d95cac011) which contains the surface processing yaml. Cylc review logs [here](http://fcm1/cylc-review/taskjobs/rnixonhi/?suite=ukv_surface_20210701T1200Z_no_openroad_no_vis_ops_rebased_jjaux%2Frun1).
Analysis_increments_Hist_lev_01

I expect that visibility will need to be added back in when that is ported.


ctgh (approved):
Thanks for adding this. I like the table in the namelist file.

ldhawknesssmith (approved):

PR #202, opened 2024-05-05
Remove literal kinds

### LINKED ISSUE #138: Integer literals used for kind values
Following on from https://github.com/MetOffice/opsinputs/pull/133#discussion_r1033564552.

This applies to all types of variables, not just logicals.  Literal kind values (4, 8, etc) are not portable between implementations although in practice values of 4 and 8 corresponding to the number of bytes are very common.  More portable is to use named constants (parameters) that might be different for different compilers.  This can be done via cpp macros or selected_*_kind.  Also more recent standards have standard constants defined in the intrinsic module iso_fortran_env.

For example, OPS's implementation of this is here: https://www-nwp/~opsrc/OPS/view/ops-2022.01.1/doc/OpsMod_Kinds.html.  This contains a lot of legacy cruft.  Unless there is a problem with one of the compilers I would try just using iso_fortran_env constants.

DJDavies2:
Resolves #138.  I don't think there are any other cases.  c_int seems most consistent with everything else.

Test output:

http://fcm1/cylc-review/taskjobs/frwd?&suite=opsinputs-138

### DIFF
--- a/src/opsinputs/opsinputs_fill_mod.F90
+++ b/src/opsinputs/opsinputs_fill_mod.F90
@@ -1782,9 +1782,9 @@ character(len=*), intent(in)                       :: JediVarName
 character(len=*), intent(in)                       :: JediVarGroup
 
 ! Local declarations:
-integer(kind=4)                                    :: VarValue(JediToOpsLayoutMapping % NumJediObs)
-integer(kind=4)                                    :: CurrentVarValue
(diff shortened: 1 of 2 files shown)


mikecooke77 (approved):
Thanks for making this change

ctgh (approved):
Thanks, looks good!

PR #203, opened 2024-05-05
AOCC support

DJDavies2:
This is some minimal changes for AOCC porting on EXZ.  Test output:

http://fcm1/cylc-review/taskjobs/frwd?&suite=opsinputs-201

### DIFF
--- a/deps/CMakeLists.txt
+++ b/deps/CMakeLists.txt
@@ -37,6 +37,12 @@ elseif ("${CMAKE_Fortran_COMPILER_ID}" MATCHES "Intel")
 ${CMAKE_Fortran_FLAGS}"
 #-Wp,-P \
   )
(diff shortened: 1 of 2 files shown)


mikecooke77 (approved):
Looks good thanks for adding.  Just for my own interest are the gnu settings portable between the old and new hpc?

ctgh (approved):

yaswant (approved):

DJDavies2:
> Looks good thanks for adding. Just for my own interest are the gnu settings portable between the old and new hpc?

I haven't needed to change them.  However I don't know whether they are optimal or not.

It would be better if compile options like these could be set outside of the source.  It isn't ideal to have to make code changes like this to tweak compile options.

PR #206, opened 2024-05-13
Account for integer `MetaData/stationIdentification`

### LINKED ISSUE #205: Account for integer `MetaData/stationIdentification`
The radar ODB mapping requires `MetaData/stationIdentification` to be mapped to an integer (the radar site number); `statid` is empty so cannot be used. This causes problems in the Varobs Writer which is hard-coded to expect a string value. Add a parameter that instructs the Varobs Writer to expect an integer instead.

ctgh:
Enable `MetaData/stationIdentification` to be an integer variable. This is required for radar processing, which maps it to the integer site ID. (`statid` is empty so cannot be used.)

By default this is not performed - the user has to add a yaml option to enable it.

This PR includes a unit test of the new feature.

Fixes #205


ReubenHill (approved):
Whilst I'm not entirely sure of the exact limits of what you can and cannot do with Fortran 90, this all looks fine to me.

JoanneAWaller (approved):

matthewrmshin:
Is this ready to merge?

ctgh:
@matthewrmshin Yes thanks!

ctgh:
Thanks @ReubenHill and @JoanneAWaller for your reviews.

PR #208, opened 2024-05-21
Update global surface CX namelist

### LINKED ISSUE #155: Use explicit CX namelist for global surface observations

ctgh:
Explicitly list the fields in the global surface CX namelist. This follows the UKV namelist but adds three more required fields: cloud_bulk, cf and cl.

The existing opsinputs ctest is unchanged.

Sith global KGO output: http://fcm1/cylc-review/taskjobs/chthomas/?suite=sith-cx-global%2Frun3

(Not shown) sith UKV and malak global KGOs are also unaffected.

Fixes #155


JoanneAWaller (approved):
Thanks @ctgh for making the cx list explicit, it is much clearer than before.

etc/global/cx/Surface.nl
```diff
@@ -1,5 +1,49 @@
 &CXControlNL
-cxfields_remove=23, 25, 31, 90, 1235, 3208, 3247, 3255, 3464, 3465, 8223, 8225, 9203, 9204, 9217, 9219, 16222, 20000, 20001, 20002
-cxfields_add=266, 267, 268
-
+! All variables below are found in OPS cx outputs for surface processing. Not all are
+! necessary for JOPA.
```
ReubenHill, line 3: Has this actually been checked in this case? If not I think the comment should be amended. The list is useful, I just don't want us to imply we've confirmed these are the fields that are output by OPS when we haven't actually and cause confusion down the line!

ReubenHill (commented):
Excellent! Just one comment to address

etc/global/cx/Surface.nl
ctgh: Good point - taking a look...

ctgh: I have updated the table in be4a150. There were a few changes to make in the surface variables, and nothing for the upper-air ones.

ReubenHill (approved):
Thanks!

ctgh:
Thanks for both of your reviews!

PR #212, opened 2024-06-24
Add files for radar Doppler wind processing

### LINKED ISSUE #211: Add files for radar Doppler wind processing
Add files for radar Doppler wind processing

ctgh:
Add files for radar Doppler wind processing:
* CX and Varobs namelists.
* UKV unit tests for the CX and VarObs writers.
* Varobs unit test for radial velocity (varobs index 63). I removed the test for radar azimuth (index 66) because it is not appropriate. Due to the setup of the VarObs writer, when radial velocity files are created they automatically include indices 64, 65, 66, 69 and 75.
* Addition of several radar variables to opsinputs_varobswriter_mod.F90. These auxiliary variables are put into the VarObs data structure as described above.
* Addition of the function `opsinputs_fill_fillinteger2d` in opsinputs_fill_mod.F90. This enables the variable `MetaData/stationIdentification` (which in this case is mapped to an integer rather than a string) to be processed correctly. NB 1: a 2D function is required here due to the OPS data structure. NB 2: I did not copy all of the functionality in `opsinputs_fill_fillreal2d` to this function. Instead, that can be added as necessary. Given opsinputs is quite mature I suspect that may not be required.

Fixes #211


etc/ukv/cx/Radar.nl
```diff
@@ -0,0 +1,3 @@
+&CXControlNL
+CxFields=2,3,33,150
+/
```
ReubenHill: Comments here, and in the varobs namelist, might be nice to say exactly which fields these are.

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1817,6 +1818,87 @@ subroutine opsinputs_fill_fillinteger( &
...
+!>   Indices returned by ioda::ObsSpace::obsvariables().channels().
+!> \param[in] VarobsLength
+!>   Length of varobs profile.
+!> \param[in] JediVarName
+!>   Name of the JEDI variable used to populate \p Int2. If each JEDI location needs to be mapped
+!>   to a separate OPS observation, this can represent either a single variable with no channel
+!>   suffix (in which case \p Int2 will have only a single row) or a set of variables with
+!>   suffixes corresponding to the indices specified in \p Channels.
+!> \param[in] JediGroup
+!>   Group of the JEDI variable used to populate \p Int2.
```
ReubenHill: Should this say `Group name of the ...`?

test/generate_unittest_netcdfs.py
```diff
@@ -34,9 +34,27 @@ def output_1d_simulated_var_to_netcdf(var_name, file_name, with_bias=False):
     minute = 1 / 60.
     var[:] = [1 * minute, 2 * minute, 3 * minute, 4 * minute]
 
-    var = f.createVariable('MetaData/stationIdentification', str, ('Location'))
-    for i, s in enumerate(["station_1", "station_2", "station_3", "station_4"]):
-        var[i] = s
+    # Station ID
+    if radar_doppler_wind:
+        var = f.createVariable('MetaData/stationIdentification', 'i', ('Location'))
+        var[:] = [1, 2, 3, 4]
```
ReubenHill: Comment here that radar doppler wind uses integers for `stationIdentification`? And give the reason perhaps?

src/opsinputs/opsinputs_fill_mod.F90
ReubenHill: And the same in the other subroutines

ReubenHill (commented):
Thanks for this. I just have a few minor comments for you to respond to then I'll be happy to approve.

etc/ukv/varobs/Radar.nl
```diff
@@ -0,0 +1,3 @@
+&VarobsControlNL
+Varfields=63,64,65,66,69,75
```
PJLevensMO, line 9: I'm not sure I fully understand why varfield 66 is requested here but the test for it has been removed below - is that because it is used in the processing but does not need to be passed to VAR?

PJLevensMO (commented):
Hi Chris, just one small question from me. I've not run this through JOPA but I trust you have and it works as intended.

ctgh:
Thanks for both of your reviews! I will respond today.

etc/ukv/varobs/Radar.nl
ctgh: This is a quirk of the Varobs structure that is used for Doppler winds. 63 is the observable of interet (radial wind), and the other ones are all metadata. OPS outputs all of these at once. Therefore I thought it made more sense to add a ctest that does this, rather than focus on the individual metadata variables. Furthermore, the ctest for variable 66 wasn't working because the obsspace was called `Radar` (and therefore expected all of the auxiliary variables to be present). There are probably ways around that but I don't think they would add much beyond the existing ctest.

PJLevensMO: Thanks for the answer Chris, that makes sense! I'll resolve this conversation now, happy to approve

ctgh:
@ReubenHill I have implemented all of your suggestions in a662ba6.

PJLevensMO (approved):
Thanks for addressing all of the comments, Chris. Looks good to me.

ctgh:
In 591e7f4 I updated deps/ops/stubs/OpsMod_Varobs/Ops_VarobPGEs.inc in order to write out the radial wind PGEs correctly. The main issue with the existing code is that the logical variable `RadWind_SuperOb` will never be `true` in opsinputs (whereas it is `true` in OPS). I decided to remove the test on that variable and ensure that the correct PGE is written out.

@mikecooke77 As mentioned in the comment above, I have modified one of the OPS stubs because it was leading to inconsistent behaviour relative to the actual OPS code. This change should not be ported into OPS itself. Do you foresee any issues with doing that?

mikecooke77:
> In [591e7f4](https://github.com/MetOffice/opsinputs/commit/591e7f4011be080f92afe5aeb2e27ed636dd80d0) I updated deps/ops/stubs/OpsMod_Varobs/Ops_VarobPGEs.inc in order to write out the radial wind PGEs correctly. The main issue with the existing code is that the logical variable `RadWind_SuperOb` will never be `true` in opsinputs (whereas it is `true` in OPS). I decided to remove the test on that variable and ensure that the correct PGE is written out.
> 
> @mikecooke77 As mentioned in the comment above, I have modified one of the OPS stubs because it was leading to inconsistent behaviour relative to the actual OPS code. This change should not be ported into OPS itself. Do you foresee any issues with doing that?

The issue comes when you try and update the code for changes to the OPS e.g. we have a new datatype.  If you copy straight from OPS changes like this will disappear.  At one point we were thinking of having some sort of "copy fixes back in" kind of functionality but I don't think this is ever used.

ctgh:
> > In [591e7f4](https://github.com/MetOffice/opsinputs/commit/591e7f4011be080f92afe5aeb2e27ed636dd80d0) I updated deps/ops/stubs/OpsMod_Varobs/Ops_VarobPGEs.inc in order to write out the radial wind PGEs correctly. The main issue with the existing code is that the logical variable `RadWind_SuperOb` will never be `true` in opsinputs (whereas it is `true` in OPS). I decided to remove the test on that variable and ensure that the correct PGE is written out.
> > @mikecooke77 As mentioned in the comment above, I have modified one of the OPS stubs because it was leading to inconsistent behaviour relative to the actual OPS code. This change should not be ported into OPS itself. Do you foresee any issues with doing that?
> 
> The issue comes when you try and update the code for changes to the OPS e.g. we have a new datatype. If you copy straight from OPS changes like this will disappear. At one point we were thinking of having some sort of "copy fixes back in" kind of functionality but I don't think this is ever used.

I see. Presumably I could a test to cover this eventuality (i.e. check the output PGEs are as expected). If the file is ever overwritten the test will start to fail. Do you know if this version of OPS code has been updated?

Failing that, I might be able to hack things in the varobswriter. I'll take a look at that too.

ctgh:
In fa19be4 I removed the modification to the OPS code. Instead both `Ob % RadialVelocSO` and `Ob % RadialVelocity` are filled. It's a bit hacky but causes the PGE to be filled correctly. The unit tests pass, but I am also checking the full end-to-end test. I will let you know how it goes...

ctgh:
> In [fa19be4](https://github.com/MetOffice/opsinputs/commit/fa19be4cc85b24ca022e79467eb70c1d9338ebd0) I removed the modification to the OPS code. Instead both `Ob % RadialVelocSO` and `Ob % RadialVelocity` are filled. It's a bit hacky but causes the PGE to be filled correctly. The unit tests pass, but I am also checking the full end-to-end test. I will let you know how it goes...

I can confirm that this change does not affect the PGE output in a full test.

mikecooke77 (approved):
Nothing to add it all looks good to me.  I think removing the changes from the dependcies will be helpful.

ReubenHill (approved):

ctgh:
Thanks all for your reviews!

PR #216, opened 2024-07-12
Adding windspeed varobs test and oceanwinds var_cx test

ghalloran:
For processing Oceanwinds observations (ocean surface windspeed observations from eg.GNSS-reflectometry) we need to add var_cx capability and windspeed to the varobs.


test/testinput/cxwriter_globalnamelist_oceanwinds.yaml
```diff
@@ -0,0 +1,41 @@
...
+      - variable:
+          name: MetaData/latitude
+        minvalue: 0.0
+    - filter: Cx Writer
+      namelist_directory: ../etc/global/cx
+      reject_obs_with_any_variable_failing_qc: true
+      general_mode: debug
+      IC_PLevels: 5
+    - filter: Cx Checker
+      expected_surface_variables: ["1","2","5","6","13","17"] # "16"
```
james-cotton: Very minor tidy - can remove variable 16 which is commented out

james-cotton (approved):
Hi Gemma, the changes you have made look all good to me. Thanks for adding both the varobs and cxwriter tests too!

ctgh (commented):
Thanks for adding this - everything looks good. Do you also need to add a namelist for the varobs writer (i.e. etc/global/varobs/Oceanwinds.nl)?

ghalloran:
> Thanks for adding this - everything looks good. Do you also need to add a namelist for the varobs writer (i.e. etc/global/varobs/Oceanwinds.nl)?

Good question - I don't know! This doesn't include a test for an oceanwinds.varobs file itself, just adds the capability to output the surface windspeed itself, so I don't think it needs a namelist. I was just following the instructions on the opsinputs home page for the varobs part, and then for the cx part the instructions didn't help, so followed what has been done before for other obtypes.

ctgh:
Thanks for explaining. Eventually you will need a varobs namelist in order to send output from JOPA to VAR. However I am happy to wait for a new PR to do that if you prefer.
Sorry the instructions are out of date - copying what has been done previously is usually the best option.

james-cotton:
If you don't supply a varobs namelist it will just use the code defaults, e.g. https://github.com/MetOffice/opsinputs/blob/develop/deps/ops/stubs/OpsMod_Varobs/Ops_GetDefaultVarfields.inc#L214
In this case we are only ever outputting a single varfield so the namelist isn't necessary.

ctgh:
> If you don't supply a varobs namelist it will just use the code defaults, e.g. https://github.com/MetOffice/opsinputs/blob/develop/deps/ops/stubs/OpsMod_Varobs/Ops_GetDefaultVarfields.inc#L214 In this case we are only ever outputting a single varfield so the namelist isn't necessary.

OK, thanks. Part of me would suggest adding one anyway, just to be explicit. However that could be done later if need be.
Given that, I will approve.

ctgh (approved):

PR #218, opened 2024-07-23
General Mode in post filters now as per yaml

### LINKED ISSUE #217: GeneralMode being set by the last filter
If both varobs and cx writers are being used the GeneralMode last set by the constructor is being used for both filters after this.  This should not be the case.  I think this is the case because the filters are compiled together and so make use of the same Gen modules.  The perils of passing variables through by modules.

mikecooke77:
This PR fixes #217 

In order to see how this affects the value of GeneralMode in the post filters there is a ctest in this branch.  There is no need to commit this ctest as it is not actually testing anything just adding a couple of print statements which need manually checking.

https://github.com/MetOffice/opsinputs/tree/feature/separateGenMode_withctest

### DIFF
--- a/src/opsinputs/opsinputs_cxwriter_mod.F90
+++ b/src/opsinputs/opsinputs_cxwriter_mod.F90
@@ -157,6 +157,7 @@ private
   integer(integer64)                     :: IC_BLevels
   integer(integer64)                     :: IC_WetLevels
   integer(integer64)                     :: IC_FirstConstantRhoLevel
+  integer                                :: GeneralMode ! Local scope to ensure it propogates correctly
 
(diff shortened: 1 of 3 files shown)


ctgh (approved):
Thanks for catching this. I assume your extra ctest works as expected?

mikethurlow (approved):
Happy with this. Thanks Mike

mikecooke77:
> Thanks for catching this. I assume your extra ctest works as expected?

Yes I tested it initially and it picked up the final filter (cx) number and used this for both.  I think this is because the code is compiled together and there is only one version of the Gen code and the flag is passed around by modules.  With this change the first filter (varobs) had a different value to the second filter (cx).

PR #221, opened 2024-08-15
Update varobs and cx writers for visibility

ReubenHill:
The changes made in this PR are necessary for the assimilation of visibility by VAR. There are three key changes:
1. The surface namelist has been updated to include model fields for visibility, surface total water (`qt2`) and aerosol (specifically with stash code 90, called "Total Aerosol (for Vis)").
2. The cx writer has been updated to allow surface total water and aerosol to be output. These are used by VAR, alongside other model variables, to calculate background visibility.
3. The varobs writer will now write `ObsValue/horizonalVisibility` values to `VarField_logvis`.

Updating the cx writer for outputting aerosol was non-trivial:
1. There are multiple aerosol stash codes - `90` is the correct one to use for the calculation of background visibility in VAR.
2. The field is defined over all 70 levels, but OPS only outputs the level closest to the surface: see the comments in `src/opsinputs/opsinputs_fill_mod.F90` for details. This required special case code.

@adammaycock this was the code you were helping me to debug last month: the issue was an incorrect stash code causing the necessary select case to be skipped over.

@ss421 I have now confirmed that I need the following for VAR to run, so please don't exclude them from [FieldNameCompare.xlsx](https://metoffice.sharepoint.com/:x:/r/sites/metofficengmsprogrammeext/NGPAO/ModelVariableNames/FieldNameCompare.xlsx?d=wdb1aad32a2a947cdaf35389e5a492ae7&csf=1&web=1&e=chZEHJ):
 - qt2 - surface total water - stash code 3255
 - aerosol - Total Aerosol (for Vis) - stash code 90

**NOTE**: the ukv surface KGO will need to be remade once this is merged.


src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1380,17 +1382,33 @@ subroutine opsinputs_fill_fillrealfromgeoval( &
...
+    if (JediVarName == "aerosol") then
+      ! This field "Total Aerosol (for Vis)" exists on 70 levels, but OPS only
+      ! outputs (and VAR expects) a single value at the surface (GeoVal level
+      ! 70 or cx level 1) for the calculation of visibility. See
+      ! Ops_BGEandCXCreate.inc line 166 and Ops_CxComplete.inc line 299. This
+      ! field is only used for surface visibility calculations so no further
+      ! specific handling is required.
+      GeoValShape = shape(GeoVaL % vals)
+      NumLevs = GeoValShape(1)
+      where (GeoVal % vals(NumLevs,:) /= MissingReal)
```
ctgh: To avoid duplicating this `where` block, perhaps you could set the level index in an `if-else` block.
```
if (JediVarName == "aerosol") then
  GeoValShape = shape(GeoVaL % vals)
  LevelIndex = GeoValShape(1)
else
  LevelIndex = 1
end if

where (GeoVal % vals(LevelIndex,:) /= MissingReal)
  Real1 = GeoVal % vals(LevelIndex,:)
end where
```

ctgh (commented):
Very nice! Does this require any modifications to sith KGO output?

ReubenHill:
> Very nice! Does this require any modifications to sith KGO output?

I don't think so: don't KGOs only use the NetCDF output from the jopa background task? This ought not to effect that.

ctgh:
I believe they check ODB, Varobs and CX output. It might be worth running it just to be on the safe side.

ReubenHill:
> I believe they check ODB, Varobs and CX output. It might be worth running it just to be on the safe side.

Yes you are correct, the kgos will need to be remade. Unless it's very straightforward, it's probably only worth doing once the jjdocs PR is ready since that will also need new kgos.

src/opsinputs/opsinputs_fill_mod.F90
ReubenHill: Done in b0f5598

ctgh (approved):
Thanks for making the change. The ctest failure is fixed in #222

matthewrmshin (approved):
Ran OK in my environment. No failures detected for mo-bundle main build + ctest, and sith/malak kgo runs (apart from what's expected).

ReubenHill:
OFFICIAL

I guess we don't have a UKV surface KGO set up then. Please merge if you are happy.


OFFICIAL

________________________________

matthewrmshin:
OFFICIAL

Hi Reuben,

I just realised my typing was cut off halfway. I actually meant that there were no unexpected KGO issues. I've now amended my comment in the PR. Sorry for the confusion.

Best,
Matt
---
Dr Matt Shin  Expert Scientific Software Engineer
Met Office  FitzRoy Road  Exeter  EX1 3PB  United Kingdom
Tel: +44 (0)330135 1724
***@***.***  http://www.metoffice.gov.uk/


OFFICIAL

________________________________

src/opsinputs/opsinputs_fill_mod.F90
```diff
@@ -1380,16 +1382,31 @@ subroutine opsinputs_fill_fillrealfromgeoval( &
...
-      Real1 = GeoVal % vals(1,:)
+    if (JediVarName == "aerosol") then
+      ! This field "Total Aerosol (for Vis)" exists on 70 levels, but OPS only
+      ! outputs (and VAR expects) a single value at the surface (GeoVal level
+      ! 70 or cx level 1) for the calculation of visibility. See
+      ! Ops_BGEandCXCreate.inc line 166 and Ops_CxComplete.inc line 299. This
+      ! field is only used for surface visibility calculations so no further
+      ! specific handling is required.
+      GeoValShape = shape(GeoVaL % vals)
+      LvlIdx = GeoValShape(1)  ! Usually 70
```
DJDavies2, line 1403: This can be done like this:

LvlIdx = size(GeoVaL % vals, dim=1)

DJDavies2 (approved):

src/opsinputs/opsinputs_fill_mod.F90
ReubenHill: Thanks! Changed in b4efa39

ReubenHill: I take it back - this makes the build fail. Is the syntax something different? I will revert the change for now.

ReubenHill: I take it back - this makes the build fail. I will revert the change.

matthewrmshin:
This change appears to require KGO changes for various `Surface.*` files for sith `kgo_glu` and `kgo_ukv` as well as malak `kgo_glu_jopa_jada` and `kgo_glu_jopa_var` - so not just the UKV ones. Is this expected?

ReubenHill:
> This change appears to require KGO changes for various `Surface.*` files for sith `kgo_glu` and `kgo_ukv` as well as malak `kgo_glu_jopa_jada` and `kgo_glu_jopa_var` - so not just the UKV ones. Is this expected?

The varobs writer now outputs visibility - since these are in the list of observed variables for both global and ukv yamls, I therefore expect both sets of varobs files to change. The changes here should not effect global surface cx files, though I can imagine that a change to a varobs file might somehow necessitate a change to a cx file. What exact differences do you see?

matthewrmshin:
Differences reported for these files:

malak:
* kgo_glu_jopa_jada:
  * glu_jopa_varobs/Surface.varobs
* kgo_glu_jopa_var:
  * glu_jopa_varobs/Surface.varobs

sith:
* kgo_glu:
  * glu_jopa_varobs/Surface.varobs
  * glu_jopa_varobs_screen/Surface.varobs
* kgo_ukv: 
  * ukv_jopa_var_cx/Surface.cx
  * ukv_jopa_varobs/Surface.varobs

See outputs of these suites of mine: http://fcm1/cylc-review/suites?user=frsn&names=*oi221-kgo_glu*+*oi221-kgo_ukv

ReubenHill:
> Differences reported for these files:
> 
> malak:
> 
> * kgo_glu_jopa_jada:
>   
>   * glu_jopa_varobs/Surface.varobs
> * kgo_glu_jopa_var:
>   
>   * glu_jopa_varobs/Surface.varobs
> 
> sith:
> 
> * kgo_glu:
>   
>   * glu_jopa_varobs/Surface.varobs
>   * glu_jopa_varobs_screen/Surface.varobs
> * kgo_ukv:
>   
>   * ukv_jopa_var_cx/Surface.cx
>   * ukv_jopa_varobs/Surface.varobs
> 
> See outputs of these suites of mine: http://fcm1/cylc-review/suites?user=frsn&names=*oi221-kgo_glu*+*oi221-kgo_ukv

These are as expected. Whilst the global varobs files now contain visibility (as the OPS ones always have), the global JOPA VAR task `job.stats` outputs [before](http://fcm1/cylc-review/view/rnixonhi?&suite=gl_surface_20210701T1200Z_no_opsinputs_update%2Frun1&no_fuzzy_time=0&path=log/job/20210701T1200Z/glu_var_jopa_anal_low/01/job.stats) and  [after](http://fcm1/cylc-review/view/rnixonhi?&suite=gl_surface_20210701T1200Z_opsinputs_update%2Frun1&no_fuzzy_time=0&path=log/job/20210701T1200Z/glu_var_jopa_anal_low/01/job.stats) are identical - i.e. the presence of visibility in the global surface JOPA output has no impact on VAR. So please go ahead and recreate the KGOs.

PR #230, opened 2024-10-10
Model variable naming sprint: part 2

ss421:
## Description

As part of variable name code sprint, this PR renames variables from group 2:

`specific_humidity` -> `water_vapor_mixing_ratio_wrt_moist_air_and_condensed_water`
`specific_humidity_at_two_meters_above_surface` -> `water_vapor_mixing_ratio_wrt_moist_air_and_condensed_water_at_2m`

## Dependencies

build-group=https://github.com/JCSDA-internal/vader/pull/278
build-group=https://github.com/JCSDA-internal/saber/pull/956
build-group=https://github.com/JCSDA-internal/ufo/pull/3489
build-group=https://github.com/JCSDA-internal/ufo-data/pull/455
build-group=https://github.com/JCSDA-internal/ops-um-jedi/pull/238
build-group=https://github.com/JCSDA-internal/lfric-jedi/pull/915
build-group=https://github.com/MetOffice/jjtests/pull/53
build-group=https://github.com/JCSDA-internal/oops/pull/2767

## Impact

No impacts expected beyond the list in dependencies above

## Checklist

- [x] I have performed a self-review of my own code
- [ ] I have made corresponding changes to the documentation
- [ ] I have run the unit tests before creating the PR

PR #231, opened 2024-10-14
Changes to remove _and_condensed_water from specific humidty

mikecooke77:
Given the change to move to a variable transform in the lfric-jedi variable change factory is accepted then this change will be needed in opsinputs.


DJDavies2 (approved):

ss421 (approved):
Thankyou.

ss421:
Given this will be fully tested as part of PR: #230, merging...

PR #234, opened 2025-02-07
Add capability to create varobs and cx files for MTG-IRS 

stemiglio:
Add MTG-IRS ObsGroup so that `OpsMod_Control` need to be moved under `stubs` to avoid removing the change when updating OPS. Also add necessary code to allow the creation of MTG-IRS varobs and the `MTGIRS.nl` namelists to for MTG-IRS varobs and cx files


ctgh:
Thanks for adding this. Can you please add ctests for the varobs and CX writers?

DJDavies2 (approved):

stemiglio:
> Thanks for adding this. Can you please add ctests for the varobs and CX writers?

Good point, thanks for reminding me. I have now added `test_opsinputs_cxwriter_globalnamelist_mtgirs` and `test_opsinputs_varobswriter_globalnamelist_mtgirs` ctests and I can confirm they succeed after compilation

ctgh (approved):
Thanks for adding the tests! I'm approving but I'd like to double check about the change in OPS code location. @adammaycock or @mikecooke77 can I check what you think please?

mikecooke77 (changes requested):
All of the test files used with opsinputs should be created using the https://github.com/MetOffice/opsinputs/blob/develop/test/generate_unittest_netcdfs.py
script.  This allows for reproducibility.  This will result in changes to the ctest output which will need to be modified.

stemiglio:
Hi @mikecooke77, thanks for spotting the missing updated python file. I can confirm that I used the python script to generate the MTG-IRS test files, but I forgot to add the modified version to this PR. Now added. Note I tested that the varobs and cx files produced by the script are the same as those associated with this PR.

mikecooke77:
> Thanks for adding the tests! I'm approving but I'd like to double check about the change in OPS code location. @adammaycock or @mikecooke77 can I check what you think please?

I think this is the correct solution.

mikecooke77 (approved):

ctgh:
@stemiglio When you're ready, please add the 'ready to merge' label and one of us can merge this.

ctgh:
There is a CI failure but it seems to be something to do with python?

stemiglio:
Yes, I can see that. I'll do an empty commit to see if it goes away

stemiglio:
Nope, there is still this error:
CMake Error at ioda/src/engines/ioda/CMakeLists.txt:344 (target_link_libraries):

  Target "ioda_engines" links to:

    Python3::Python

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

@matthewrmshin @mikecooke77 do you know what's the cause of it?

stemiglio:
@matthewrmshin I have now tried your suggestion and as you can see, with the older version of the JCSDA container the problem disappear. What do you suggest we do now?

stemiglio:
After liasing with @matthewrmshin I have now created https://github.com/MetOffice/opsinputs/issues/236. For now we can use the previous container image.

ctgh:
Thanks, merging now!

ctgh:
Matt beat me to it! 😆

PR #235, opened 2025-02-07
Add FCI on Meteosat-12 varobs and CX namelists and test files

PJLevensMO:
This PR adds the files needed for processing FCI data in the UKV, as well as updating the inbuilt OPS routines to handle FCI/Met-12 data.


DJDavies2 (approved):

smnewman:
Hi @PJLevensMO, are these changes in `deps/ops` being mirrored in OPS?

PJLevensMO:
> Hi @PJLevensMO, are these changes in `deps/ops` being mirrored in OPS?

The changes in OPS are tracked in this ticket: https://code.metoffice.gov.uk/trac/ops/ticket/2961

It has been a while since I've compared the two, I will check that everything covered here is the same in the OPS branch

PJLevensMO:
@smnewman I've found a couple of files that are either missing updates vs. OPS or could be tidied up. I'll just edit those and push the changes.

smnewman (approved):
Looks good to me @PJLevensMO. I assume it's not possible to test the code in sith until we have real data?

PJLevensMO:
> Looks good to me @PJLevensMO. I assume it's not possible to test the code in sith until we have real data?

We do have real data now, I ran some tests ages ago on the test data which worked fine. I can try and set up a sith test which uses the operational data just to test the code works, if that would be useful?

smnewman:
> > Looks good to me @PJLevensMO. I assume it's not possible to test the code in sith until we have real data?
> 
> We do have real data now, I ran some tests ages ago on the test data which worked fine. I can try and set up a sith test which uses the operational data just to test the code works, if that would be useful?

I think some sith output would be a useful sanity check, if feasible.

PJLevensMO:
I'll get that set up. With the caviat that this would purely be a technical test - the output might be nonsense (R matrix and biases have not been tuned for FCI, so using SEVIRI as a baseline) and JOPA yamls have not been properly implemented yet.

PJLevensMO:
It has taken a little while to get sith working, there is a problem with the VarBC file I had been using (unrelated to this PR). Removing that, however, the opsinputs parts work as expected. See the output at http://fcm1/cylc-review/taskjobs/plevens/?suite=fci_test%2Frun19

mikecooke77:
I have just deleted the files and regenerated locally before running the ctests and all ctests pass.  This will not affect KGOs and is therefore being merged.

PR #238, opened 2025-03-07
Fix ifort truncated_source warning to error issue

matthewrmshin:
Explicitly list warn types, instead of all.

### DIFF
--- a/deps/CMakeLists.txt
+++ b/deps/CMakeLists.txt
@@ -23,7 +23,7 @@ ${CMAKE_Fortran_DEBUG_FLAGS}"
   )
 elseif ("${CMAKE_Fortran_COMPILER_ID}" MATCHES "Intel")
   set(OPS_DEFAULT_Fortran_FLAGS "\
--warn alignments,errors,declarations,general,ignore_loc,nointerfaces,notruncated_source,shape,stderrors,uncalled,unused,usage \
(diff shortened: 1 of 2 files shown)


deps/CMakeLists.txt
```diff
@@ -23,10 +23,7 @@ ${CMAKE_Fortran_DEBUG_FLAGS}"
   )
 elseif ("${CMAKE_Fortran_COMPILER_ID}" MATCHES "Intel")
   set(OPS_DEFAULT_Fortran_FLAGS "\
--warn all \
--warn errors \
--warn notruncated_source \
--warn nointerfaces \
+-warn alignments,errors,declarations,general,ignore_loc,nointerfaces,notruncated_source,shape,stderrors,uncalled,unused,usage \
```
ctgh: Tiny point: you could swap `errors` and `declarations` to put it in alphabetical order.

ctgh (approved):
Thanks for this fix. I wasn't familiar with all of the options you can use here, but they seem to match the list on this site: https://www.cita.utoronto.ca/~merz/intel_f10b/main_for/mergedProjects/copts_for/fortran_options/option_warn.htm

deps/CMakeLists.txt
matthewrmshin: Done.

DJDavies2 (approved):

PR #246, opened 2025-08-07
Add cx , varobs namelists for EPSMWS obsgroup

brettcandy:
This change adds namelist files specifying cx and varobs fields for obsgroup EPSMWS (mw sounder on MetopSG). These are identical to SternaMWS in terms of content. I also noticed that the specification for default varobs fields for both Sterna and EPS were incorrect - the bias predictors are specified twice.


brettcandy:
**Tests in sith** 

I have built the code on the EX matchines and have run a single cylc of my sith branch for EPSMWS. Change performs as expected. see output [here](https://cylchub/services/cylc-review/taskjobs/brett.candy?&suite=sith_epsmws&cycles=20240912T0600Z)

brettcandy:
This is now ready for review.

DJDavies2 (approved):

deps/ops/stubs/OpsMod_Varobs/Ops_GetDefaultVarfields.inc
```diff
@@ -285,15 +285,15 @@ SELECT CASE (ObsGroup)
                         VarField_tcozone,VarField_solzenith,VarField_QCinfo, &
                         VarField_clw/)
   CASE (ObsGroupEPSMWS) !setup for jedi
-    Varfields(1:10) = (/Varfield_britemp,Varfield_satid,Varfield_satzenith, &
+    Varfields(1:9) = (/Varfield_britemp,Varfield_satid,Varfield_satzenith, &
                         Varfield_solzenith,Varfield_tskin, &
                         Varfield_surface,Varfield_numchans,Varfield_channum, &
-                        Varfield_mwemiss, varfield_biaspredictors /)
```
ctgh, line 291: Just to check: you mentioned that the bias predictors are specified twice, but they only seem to appear once in each of these lists.

ctgh (commented):
Thanks for adding this. Would you be able to add a CX writer test too, please?

deps/ops/stubs/OpsMod_Varobs/Ops_GetDefaultVarfields.inc
brettcandy: > Just to check: you mentioned that the bias predictors are specified twice, but they only seem to appear once in each of these lists.

later in the routine at L301 varfield_biaspredictors are added to the array so it ends up being added twice

brettcandy:
I have added ctests for cxwriter for SternaMWS and EPSMWS in changesets 333dce2dfcd5e5aa635a77f666713bd283459972 and 8bb0c4b283c3be48ee2e89c52216beb81f2b1fcb

brettcandy:
The additional ctests are ready now and work - do you require anything else?
thanks

mikecooke77 (approved):

ctgh (approved):
Thanks for adding the CX writer test. Can you please add a 'ready to merge' label and also retrigger the CI? It looks like there was a transient problem which should disappear if it is rerun.

brettcandy:
Any idea why the CI tests are failing @matthewrmshin ?  i have retriggered several times over several days

mikecooke77:
> Any idea why the CI tests are failing @matthewrmshin ? i have retriggered several times over several days

I do wonder if we should just use our own CI rather than trying to use there image for this testing.  opsinputs is only ever going to be used in our environment.

ctgh:
@mikecooke77 Shall we go ahead and merge this, given it passes the tests on our local system?

brettcandy:
> @mikecooke77 Shall we go ahead and merge this, given it passes the tests on our local system?

yes or advise me what to do to get CI working please

brettcandy:
bump - any progress here ?

ctgh:
I'll just bypass the rules. It shouldn't affect our internal CI, but I will keep an eye on the nightlies.

brettcandy:
grazias

matthewrmshin:
Most likely because CI image is old. Just checked. The image jcsda/docker-clang-mpich-dev:latest has eckit 1.24.5 and the recent change to ioda requires a more up-to-date version of eckit.

ctgh:
@matthewrmshin Thanks for looking into that. Is there anything that can be done from our side to get that up to date?

matthewrmshin:
@ctgh Yes, (and bearing in mind that this is a public repo like orca-jedi), we can do what orca-jedi is doing and build our own container image for CI purpose and storing the image under GitHub container registry.

PR #248, opened 2025-09-15
Add FY3-E MWSFY3E obtype

david-rundle:
Add MWSFY3E obtype which is an almost straight copy of MWSFY3 with a few more channels (but it's still store in the MWTS MetDB type)


DJDavies2 (approved):

deps/ops/stubs/Ops_Constants/Ops_SubTypeNameToNum.inc
```diff
@@ -200,12 +200,14 @@ SELECT CASE (name)
...
     num = ObsTypeMWHSFY3
+  CASE ("MWHS2E")
+    num = ObsTypeMWHSFY3E
   CASE ("MWRI")
     num = ObsTypeMWRI
   CASE ("MWTS")
     num = ObsTypeMWSFY3
-  CASE ("MWSSG")
-    num = ObsTypeMWSSG
+  CASE ("MWTS3")
```
ctgh, line 211: For my information, why did you change MWSSG to MTWS3?

ctgh (commented):
Thanks for adding this. I realise it's very similar to an existing case, but perhaps consider adding another Cx Writer ctest in order to ensure it is covered.

mikecooke77 (approved):
Looks good to me.

deps/ops/stubs/Ops_Constants/Ops_SubTypeNameToNum.inc
david-rundle: Great catch Chris!

I didn't but I must have been working from an older version of this file and missed it during merge as my addition is at exactly the same place.

david-rundle: I've fixed this now.

matthewrmshin:
CI failing the same way as mo-bundle workflow running in my environment.

david-rundle:
> Thanks for adding this. I realise it's very similar to an existing case, but perhaps consider adding another Cx Writer ctest in order to ensure it is covered.

Thanks Chris - I've added a test. Hopefully correctly. But I notice that it's failing.

I haven't added anything to ioda which I notice is used here. Do I need to?

david-rundle:
> CI failing the same way as mo-bundle workflow running in my environment.

At the very least I forgot to actually run the python script to generate the unit test python which I have now done and added.

I see that ioda is a dependency but I don't have a PR to implement a test there yet. I'm in the process of adding it now.

matthewrmshin (approved):

ctgh (approved):
Thanks for making the changes and adding the new test. I'll merge once the CI has completed.

PR #250, opened 2025-09-24
CI to use self-hosted runner

matthewrmshin:
CI to use self-hosted runner with mo-bundle's custom CI container image.

Once this is approved, I'll remove the PR check requirements to rely on the self-hosted runner CI instead.

I've also removed the ready to merge label check requirement - this is not desirable as on a listing of PRs, everything will appear to fail unless they are marked as ready to merge.


ctgh:
Thanks for adding this. Do you know why the JCSDA checks have the status "Expected — Waiting for status to be reported"?

DJDavies2 (approved):

matthewrmshin:
@ctgh We'll need to turn those off in **Settings** of this repo. I'll do that once this PR has enough approvals.

ctgh (approved):
Looks good to me although I am not very familiar with this syntax.

PR #261, opened 2025-11-04
Observation callsign changes for GPSRO

liam-hall239:
We found a bug in which opsinputs did not output the rising and setting flags for GPSRO into varobs.

We added a quality flag for the phase of occultation, rising (1) or setting (0), on the end of the observation callsign and split the last 7 characters of the callsign to include this flag at the end of the callsign.

~~An example is found in `/home/users/liam.hall/desktop/opsinputs_pr/varobs.txt` starting from line `1418`.~~
The expected output is a 16 digit callsign with the last character to be either a 1 or a 0 - the phase flag.


adammaycock:
Out of interest, how has this bug not caused a problem (difference from OPS). Does VAR do anything with the rising / setting flag?

Also, I assume this change will result in kgo differences in nightly sith / malak testing, so we'll need to run those tests and generate new kgo before merging this PR. If that's the case, we'll set the `waiting for kgo update` label on this PR.

liam-hall239:
Hi Adam, it hasn't caused any actual issues anywhere. Before, opsinputs cut off the very last digit of the callsign so the rising and setting information was lost at this point and is needed for current RO work.
The work I am doing now needs this flag and so I have made changes to my own version of VAR which splits the data into the respective occultation phase and then can be plotted.

For context, please see the graphs in [ssa-desroziers-2025](https://github.com/MetOffice/ssa-desroziers-2025/tree/develop/cycle/202501/graphs/height).

src/opsinputs/GnssroStationIDMetOffice.cc
```diff
@@ -54,7 +54,22 @@ void GnssroStationIDMetOffice::compute(const ObsFilterData & in,
...
+  std::vector<int> qualityFlags;
+  in.get(Variable("MetaData/qualityFlags"), qualityFlags);
+  std::vector<int> isRising(nlocs, 2);
+  for (size_t i=0; i < nlocs; ++i) {
+    if ((qualityFlags[i] & (1 << 13)) != 0) {
+      isRising[i] = 1;
+    } else if ((qualityFlags[i] & (1 << 13)) == 0) {
+      isRising[i] = 0;
+    }
+    else {
```
ctgh: Unless I have read the code incorrectly I don't think you can ever reach this `else` block. How about setting `isRising` to `0` by default, then having a single `if` block that sets it to `1` if the first condition is met.

ctgh (commented):
Thanks for this fix. It looks like you need to add the new variable to the test netCDF file.

DJDavies2:
One of the ctests is failing for me:

test_opsinputs_varobswriter_071_VarField_bendingangle

There are also some coding norms failures.

src/opsinputs/GnssroStationIDMetOffice.cc
neill-b: In the OPS code (https://wwwspice/~opsrc/OPS/view/ops-2025.03.0/doc/Ops_SetupVarArray.html#L188) the rising / setting bit is set to 2 if the quality flag information is absent. I guess our equivalent of this would be checking if `qualityFlags` is missing data before checking the status of the bit.

ctgh: Makes sense. If there is a chance it could be missing, checking for that would be a good idea prior to checking its 13th bit. Having said that, looking at OPS, it checks if the variable exists at all - which might be equivalent to failing to retrieve it from the obsspace?

liam-hall239:
I have just made a change to the initial commit to remove the else statement and include the variable in the constructor. I have also removed any white space from the file.

I have checked the varobs file to see if the change is implemented and it has, with the last digit of the callsign only showing 1s and 0s. It is located here: `/home/users/liam.hall/desktop/opsinputs_pr/varobs_pr.txt` for cycle 20250101T0000Z if you need a copy for the ctests.

mikecooke77:
> /home/users/liam.hall/desktop/opsinputs_pr/varobs_pr.txt

I believe for the ctests you will need to recreate the file using the python script (https://github.com/MetOffice/opsinputs/blob/develop/test/generate_unittest_netcdfs.py) as part of this PR.

neill-b:
So, the change tells me that 140 files have been modified.  Surely this is not right?

liam-hall239:
I also thought that there were so many modified files.
When running the `generate_unittest_netcdfs.py` file, it gave: "**_PermissionError: [Errno 13] Permission denied: 'testinput/001_VarField_pstar.nc4'_** " so I deleted the existing file after checking my permissions and then it worked, creating this as well as modifying the rest.
The error never mentioned the other files.

neill-b:
It appears that this changeset is the rogue one: https://github.com/MetOffice/opsinputs/pull/261/commits/89ff34b0ac400b09784d617ef88f53bbf5eabcda
I'd suggest that you revert that one, then only modify the files that need changes.

liam-hall239:
So to make sure I have this right, just push the `generate_unittest_netcdfs.py` and the file that gave an error or just the unittest python file and of course `GnssroStationIDMetOffice.cc`?

neill-b:
> So to make sure I have this right, just push the `generate_unittest_netcdfs.py` and the file that gave an error or just the unittest python file and of course `GnssroStationIDMetOffice.cc`?

I would assume that the only data file to change should be `test/testinput/071_VarField_bendingangle.nc4`. I don't understand why the python code to generate the unit tests has altered the other data files, it shouldn't need to.  That might be a question for the code owner of the python.  But anyway, I'd try reverting the changes on all the other files.

ctgh:
This has happened before - it seems to occur whenever the underlying software stack is updated, which causes small differences in the netCDF files. Previously we have just updated all of the files even if they are unrelated to the PR.

src/opsinputs/GnssroStationIDMetOffice.cc
ctgh: Do you need to have a block like `if (qualityFlags[i] == missing)` (where `missing` is equal to the JEDI missing integer)?

Also, the indentation of this block doesn't look quite right.

mikecooke77:
Not sure what to make of the CI runner.  @matthewrmshin is this because we are using an older version of the JCSDA container?  Should we move this to just using the azure UKMO CI as it will only ever be run in this environment?

matthewrmshin:
CI should now be using our own container running on our own self hosted runner.

matthewrmshin:
Build is now failing with a compiler error:

```console
/var/tmp/opsinputs/pr-261/opsinputs/src/opsinputs/GnssroStationIDMetOffice.cc:59:41: error: no matching function for call to 'missingValue(const int&)'
   59 |   const int missing = util::missingValue(missing);
      |                       ~~~~~~~~~~~~~~~~~~^~~~~~~~~
In file included from /tmp/build-and-test-j5MEtK/ioda/include/ioda/ObsDataVector.h:24,
                 from /var/tmp/opsinputs/pr-261/opsinputs/src/opsinputs/GnssroStationIDMetOffice.cc:24:
```

mikecooke77 (approved):

ctgh (approved):
Thanks, looks good!

DJDavies2:
There are failures in the ctest test_opsinputs_varobswriter_071_VarField_bendingangle.

ctgh:
@liam-hall239 I think the new file doesn't contains MetaData/qualityFlags. I haven't checked it for sure but the failed ctest output indicates that may be the case.

liam-hall239:
Hi @matthewrmshin, after speaking to Chris, it looks like the CI failure may be due to the recent
mobbs update. Local build and tests pass successfully from my [workflow](https://cylchub/services/cylc-review/taskjobs/liam.hall/?suite=mo-bundle_des_test%2Frun1).

Please let me know if you're happy for the PR to be merged.

matthewrmshin:
I've just hacked the CI image, so let's try building one more time.

matthewrmshin:
I've hacked the CI build script to use Makefile not Ninja. This should work. Final try.

ctgh:
@matthewrmshin Thanks for getting it over the line.

liam-hall239:
Ahh success - Thank you all!

PR #262, opened 2025-12-09
Add HIRAS2 observation type for use in JOPA

mo-RowanMcL:
Most changes are 1-1 with CrIS or HIRAS processing already present.

I have not created a CxWriter test, is this a necessary step here?

sith run showing correct extraction: https://cylchub/services/cylc-review/taskjobs/rowan.mclaughlin/?suite=sith%2Fhiras2_auxdocs_merge


DJDavies2 (approved):

deps/ops/stubs/Ops_Constants/OpsMod_ObsGroupInfo.f90
```diff
@@ -79,9 +79,10 @@ MODULE OpsMod_ObsGroupInfo
 INTEGER, PARAMETER :: ObsGroupFCIASR         =  68 ! => FCI all-sky radiances
 INTEGER, PARAMETER :: ObsGroupFCIClr         =  69 ! => FCI clear-sky radiances
 INTEGER, PARAMETER :: ObsGroupSternaMWR      =  70 ! => AWS & Sterna microwave radiances
-INTEGER, PARAMETER :: ObsGroupEPSMWS         =  71 ! => EPS SG MWS 
+INTEGER, PARAMETER :: ObsGroupEPSMWS         =  71 ! => EPS SG MWS
 INTEGER, PARAMETER :: ObsGroupMWSFY3E        =  72 ! => FY3E microwave sounder data
-INTEGER, PARAMETER :: max_obs_group_num      =  72 ! Number of observation groups
+INTEGER, PARAMETER :: ObsGroupHIRAS2         =  74 ! => HIRAS Radiances
```
ctgh, line 84: Should this be 73?

deps/ops/stubs/OpsMod_CXGenerate/Ops_GetDefaultCxFields.inc
```diff
@@ -313,6 +313,12 @@ SELECT CASE (ObsGroup)
                        StashItem_p,StashItem_p_Surface,StashCode_u10,            &
                        StashCode_v10,StashCode_t2,StashCode_rh2,StashCode_pmsl/)
 
+  CASE (ObsGroupHIRAS2)
```
ctgh, line 316: What do you think about using a single HIRAS case block, i.e.
```
CASE (ObsGroupHIRAS, ObsGroupHIRAS2)
```
That has been done for other obs types, e.g. GIIRS just below.

The same is true for case blocks in other functions (but some of those have different treatment between HIRAS and HIRAS2, so it may not always be appropriate).

Happy to keep as is if you prefer.

ctgh (commented):
Thanks for adding this. If it's not too much trouble to add Cx and Varobs writer tests that would be good, just to make sure everything is covered.

deps/ops/stubs/Ops_Constants/OpsMod_ObsGroupInfo.f90
mo-RowanMcL: Originally it was, however I've tried to match this ObsGroup to the groupid inside of OPS.
https://code.metoffice.gov.uk/trac/ops/browser/main/trunk/src/code/ODB/sql/group_id.hh?rev=17931
Although EPSMWS doesn't seem to match, I thought it better for future changes that these are the same, so the only available space is at 74.

deps/ops/stubs/OpsMod_CXGenerate/Ops_GetDefaultCxFields.inc
mo-RowanMcL: If you think this is worthwhile for simplifying/shortening these files I'm happy to make the change, however in case these diverge in future (and selfishly for the sake of one less OPS rebuild) I would prefer to leave it as is.

ctgh: No problem to leave it as is.

mo-RowanMcL:
> Thanks for adding this. If it's not too much trouble to add Cx and Varobs writer tests that would be good, just to make sure everything is covered.

Cx and Varobs writer tests now complete.

ctgh (approved):
Thanks for adding the new tests. Looks good!

PR #264, opened 2026-01-19
Add defalut cx fields for MTG-IRS

stemiglio:
This PR adds a list of cx fields that are used as default for MTG-IRS. The mods are particularly useful when processing MTG-IRS in the UKV as for this model configuration there is no MTGIRS.nl file

### DIFF
--- a/deps/ops/stubs/OpsMod_CXGenerate/Ops_GetDefaultCxFields.inc
+++ b/deps/ops/stubs/OpsMod_CXGenerate/Ops_GetDefaultCxFields.inc
@@ -182,6 +182,11 @@ SELECT CASE (ObsGroup)
                        StashItem_qcl,StashItem_p,StashItem_p_surface,          &
                        StashCode_u10,StashCode_v10,StashCode_t2,StashCode_rh2, &
                        StashCode_pmsl/)
+  CASE (ObsGroupMTGIRS)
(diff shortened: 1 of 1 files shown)


ctgh (approved):
Looks good to me. Do you eventually need to add a UKV namelist file?

stemiglio:
> Looks good to me. Do you eventually need to add a UKV namelist file?

Thanks for approving @ctgh. I am not planning to add a UKV namelist file, as the non-default fields (cloud_bulk, cl and cf, in UM notation) are only needed in VAR by the moisture incrementing operator, which is not used in the UKV (for now at least)

DJDavies2 (approved):

mikecooke77 (approved):
Happy for this to go in.  Do you plan to add a nl to jjaux for operational use or rely on this?

stemiglio:
> Happy for this to go in. Do you plan to add a nl to jjaux for operational use or rely on this?

For the global we do already have a namelist file for MTGIRS, see https://github.com/MetOffice/opsinputs/blob/develop/etc/global/cx/MTGIRS.nl. For the UKV we don't need more fields than the default ones, this is why I decided to create a default list of fields for MTGIRS rather than create a UKV namelist with default fields (I think this should only be used when you ask for non-default fields as done in the global). This means I was not planning to add a nl for operational use in the UKV.

PR #266, opened 2026-01-23
Remove INTEGER*4

### LINKED ISSUE #265: INTEGER*4 declarations
Fortitude is flagging this (see https://github.com/MetOffice/mo-bundle/issues/867).

DJDavies2:
Fix #265.

### DIFF
--- a/deps/ops/CMakeLists.txt
+++ b/deps/ops/CMakeLists.txt
@@ -3,7 +3,7 @@
 
 if("${CMAKE_Fortran_COMPILER_ID}" MATCHES "GNU")
   if("${CMAKE_Fortran_COMPILER_VERSION}" VERSION_GREATER "10.1.0")
-    set(OPS_EXTRA_Fortran_FLAGS "-std=f2003 -fallow-argument-mismatch")
+    set(OPS_EXTRA_Fortran_FLAGS "-std=f2008 -fallow-argument-mismatch")
   else()
     set(OPS_EXTRA_Fortran_FLAGS "-std=f2003")
   endif()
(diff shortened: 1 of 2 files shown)


matthewrmshin:
I wonder if it would be more portable using `INT32` from the [ISO_FORTRAN_ENV](https://gcc.gnu.org/onlinedocs/gfortran/ISO_005fFORTRAN_005fENV.html) module?

DJDavies2:
Yes, there are much more correct things to do that this.

Or look at removing the MATMUL_ESSL branch altogether?  I don't think it is used (but then is this whole file even used?).

I don't particularly want to go down a rabbit hole at this point.

matthewrmshin (approved):

DJDavies2:
Heh, I followed up with the next warning from Fortitude and it promptly complained about the "KIND=4", suggestiong use of ISO_FORTRAN_ENV.  So maybe I should do that here.

DJDavies2:
I've updated the branch to use int32.  This requires updating the compiler flags to 2008 standard.  There is some logic in the cmake file that tests the version; we are clearly always using at least 10.1.0 now so maybe this test can simply be removed?  I have left it in for now though.

mikecooke77 (approved):

matthewrmshin (approved):

matthewrmshin:
> I've updated the branch to use int32. This requires updating the compiler flags to 2008 standard. There is some logic in the cmake file that tests the version; we are clearly always using at least 10.1.0 now so maybe this test can simply be removed? I have left it in for now though.

Yes, I believe we can remove this test, as we should be on GNU 12 or above from now on.

DJDavies2:
> > I've updated the branch to use int32. This requires updating the compiler flags to 2008 standard. There is some logic in the cmake file that tests the version; we are clearly always using at least 10.1.0 now so maybe this test can simply be removed? I have left it in for now though.
> 
> Yes, I believe we can remove this test, as we should be on GNU 12 or above from now on.

I'll open another PR for this, it probably shouldn't be done here.

PR #268, opened 2026-01-27
Remove checks for old GCC versions

### LINKED ISSUE #267: Remove cmake tests for old versions of GNU
There are tests in the cmake files for older versions of GNU for the purposes of setting compile options.  We are not using these versions now so they can be removed.

DJDavies2:
Fix #267.  Test output:

https://cylchub/services/cylc-review/taskjobs/david.davies/?suite=opsinputs-267

### DIFF
--- a/deps/ops/CMakeLists.txt
+++ b/deps/ops/CMakeLists.txt
@@ -2,11 +2,7 @@
 #
 
 if("${CMAKE_Fortran_COMPILER_ID}" MATCHES "GNU")
-  if("${CMAKE_Fortran_COMPILER_VERSION}" VERSION_GREATER "10.1.0")
-    set(OPS_EXTRA_Fortran_FLAGS "-std=f2003 -fallow-argument-mismatch")
-  else()
-    set(OPS_EXTRA_Fortran_FLAGS "-std=f2003")
-  endif()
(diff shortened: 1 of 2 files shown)


deps/ops/CMakeLists.txt
```diff
@@ -2,11 +2,7 @@
 #
 
 if("${CMAKE_Fortran_COMPILER_ID}" MATCHES "GNU")
-  if("${CMAKE_Fortran_COMPILER_VERSION}" VERSION_GREATER "10.1.0")
-    set(OPS_EXTRA_Fortran_FLAGS "-std=f2003 -fallow-argument-mismatch")
-  else()
-    set(OPS_EXTRA_Fortran_FLAGS "-std=f2003")
-  endif()
+  set(OPS_EXTRA_Fortran_FLAGS "-std=f2003 -fallow-argument-mismatch")
```
ctgh, line 5: In #266 you are using `-std=f2008` here.

ctgh (commented):
Looks good - just one suggestion.

deps/ops/CMakeLists.txt
DJDavies2: Yes I know.  They should probably be committed separately though I think.  I can resolve the conflict.

ctgh (approved):
Thanks, fine to resolve it later.

matthewrmshin (approved):

PR #274, opened 2026-02-02
Add case default

### LINKED ISSUE #273: Fortitude: select case blocks without case default
There are some select case code blocks without case default sections.  Fortitude warns about this.

DJDavies2:
Resolves #273.

This fix is semantically equivalent to the code in develop.  However it is probably not correct as looking at the relative select case blocks it looks like if the select falls through to the end and doesn't match anything it is probably an error and the code should abort.  This would be a change of behaviour and it might fail tests.  Let me know which you would prefer.

### DIFF
--- a/src/opsinputs/opsinputs_cxwriter_mod.F90
+++ b/src/opsinputs/opsinputs_cxwriter_mod.F90
@@ -773,6 +773,8 @@ do i = 1, size(CxFields)
         write (DustBinIndexStr, '(i1)') DustBinIndex
         GeoVarName = opsinputs_cxfields_dustp_start // DustBinIndexStr // opsinputs_cxfields_dustp_end
       end if
+    case default
+      continue
 
   end select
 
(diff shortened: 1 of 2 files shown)


DJDavies2:
Test output:

https://cylchub/services/cylc-review/taskjobs/david.davies/?suite=opsinputs-273

matthewrmshin (approved):

src/opsinputs/opsinputs_cxwriter_mod.F90
```diff
@@ -773,6 +773,8 @@ subroutine opsinputs_cxwriter_addrequiredgeovars(self, geovars)
         write (DustBinIndexStr, '(i1)') DustBinIndex
         GeoVarName = opsinputs_cxfields_dustp_start // DustBinIndexStr // opsinputs_cxfields_dustp_end
       end if
+    case default
```
stemiglio, line 778: Perhaps you could add a warning message before `continue`?

stemiglio (commented):
I see your point on the probable need of a fatal error when fields are not found. But as you say this may produce changes in the tests. Perhaps we could at least ad a warning? See my comment below.

src/opsinputs/opsinputs_cxwriter_mod.F90
DJDavies2: I have done this.

stemiglio (approved):
Thanks for adding the warning messages!

matthewrmshin (approved):

PR #276, opened 2026-03-10
libopsinputs to link libops in PUBLIC mode

matthewrmshin:
This should allow other projects to correctly build opsinputs with `add_subdirectory(/path/to/opsinputs EXCLUDE_FROM_ALL)`.


matthewrmshin:
This does not work.

PR #285, opened 2026-05-18
Fortitude 0.9 happiness

matthewrmshin:
Should now passes `fortitude check` fully at fortitude 0.9.0.

Fix syntax errors manually.

Add config.
- target-std="f2008" (fortitude default is "f2018")
- line-length=140 (fortitude default is 102)
- Ignore various rules in `deps/` directory. (Many of which were legacy source files.)

Then ran with `fortitude check --fix --unsafe-fixes`.

Fix remaining failures manually.

Add `fortitude check` to CI.

See also:

- JCSDA-internal/lfric-jedi#1276
- JCSDA-internal/ops-um-jedi#269


src/opsinputs/opsinputs_cxwriter_mod.F90
```diff
@@ -173,7 +174,7 @@ module opsinputs_cxwriter_mod
   real(real64), allocatable              :: EtaTheta(:)
   real(real64), allocatable              :: EtaRho(  :)
 
-  type(ufo_geovals), pointer             :: GeoVals
+  type(ufo_geovals), pointer             :: GeoVals => null()
```
matthewrmshin, line 176: This is part of type definition of `opsinputs_cxwriter` so should be OK.,

src/opsinputs/opsinputs_varobswriter_mod.F90
```diff
@@ -199,12 +200,12 @@ module opsinputs_varobswriter_mod
...
+
   !this stores the atmospheric levels we wish to pass to varobs
-  integer(c_int), allocatable :: modlevs(:) 
-  
-  type(ufo_geovals), pointer :: GeoVals
-  type(ufo_geovals), pointer :: ObsDiags
+  integer(c_int), allocatable :: modlevs(:)
+
+  type(ufo_geovals), pointer :: GeoVals => null()
+  type(ufo_geovals), pointer :: ObsDiags => null()
```
matthewrmshin, line 207: These are part of type definition of `opsinputs_varobswriter` so should be OK.

src/opsinputs/opsinputs_obsdatavector_mod.F90
```diff
@@ -95,7 +95,6 @@ end function opsinputs_obsdatavector_float_nlocs
 
 type(obs_variables) function opsinputs_obsdatavector_float_varnames(c_vec)
   !use, intrinsic :: iso_c_binding
-  use obs_variables_mod
```
matthewrmshin, line 98: Line 9 already declared `use obs_variables_mod, only: obs_variables`, which should be sufficient here.

matthewrmshin:
Test failing. I'm going to reverse some of the less safe changes.

src/opsinputs/opsinputs_obsdatavector_interface.f90
```diff
@@ -7,69 +7,69 @@
...
 
 logical(kind=c_bool) function c_opsinputs_obsdatavector_float_has(vec, variable) &
-    bind(C,name='opsinputs_obsdatavector_float_has_f')
+    bind(C,name="opsinputs_obsdatavector_float_has_f")
   use, intrinsic :: iso_c_binding, only: c_bool, c_char, c_ptr
   implicit none
 
   type(c_ptr), value                        :: vec
-  character(kind=c_char, len=1), intent(in) :: variable(*)
+  character(kind=c_char, len=1), intent(in) :: variable(:)
```
DJDavies2: This is not an equivalent transformation, these are two very different types of arrays with very different semantics.

matthewrmshin: Thanks. Yes, I realised this cannot be done, so I'm going to revert this type of change.

matthewrmshin:
Created a new branch and blasted away the old one with a better set of changes. Tested the changes in my local environment. ~Now waiting for CI.~ All good in CI environment.

src/opsinputs/opsinputs_cxwriter_mod.F90
```diff
@@ -1311,8 +1311,8 @@ subroutine opsinputs_cxwriter_unrotatewinds(self, Ob, Cx)
 real(real64), allocatable :: Vunrot(:)               ! Array for unrotated wind v component
 real(real64), allocatable :: U10unrot(:)             ! Array for unrotated wind u10 component
 real(real64), allocatable :: V10unrot(:)             ! Array for unrotated wind v10 component
-logical                   :: UpperWinds   = .false.  ! Upper air wind u and v components present
-logical                   :: SurfaceWinds = .false.  ! Surface wind u and v components present
+logical, save             :: UpperWinds   = .false.  ! Upper air wind u and v components present
```
DJDavies2, line 1314: SAVE is redundant.

src/opsinputs/opsinputs_mpl_mod.F90
```diff
@@ -6,6 +6,7 @@
 module opsinputs_mpl_mod
 
 implicit none
+public
```
DJDavies2, line 9: This is redundant.

src/opsinputs/opsinputs_varobswriter_mod.F90
```diff
@@ -715,8 +715,8 @@ subroutine opsinputs_varobswriter_populateobservations( &
 integer                                               :: iVarField
 integer                                               :: iobs
 character(len=200)                                    :: varname
-logical                                               :: FillChanNum = .false.
-logical                                               :: FillNumChans = .false.
+logical, save                                         :: FillChanNum = .false.
```
DJDavies2, line 718: save is redundant.

DJDavies2 (approved):
The tests pass.  I am not very happy with the save and public stuff, as these are redundant and simply re-state the defaults.  I know people say they like this for "clarity", but it doesn't seem useful to me: it confuses me when I see this sort of thing because i wonder why they have been stated, and if people see both

integer :: foo = 1

and

integer, save :: foo = 1

around they might start to think there is a difference between the two when there isn't.

matthewrmshin:
Note: agree that some of the declarations may be redundant, but they allow us to satisfy these rules:

- C081 https://fortitude.readthedocs.io/en/stable/rules/initialisation-in-declaration/
- C131 https://fortitude.readthedocs.io/en/stable/rules/missing-accessibility-statement/

DJDavies2:
> Note: agree that some of the declarations may be redundant, but they allow us to satisfy these rules:
> 
>     * C081 https://fortitude.readthedocs.io/en/stable/rules/initialisation-in-declaration/
> 
>     * C131 https://fortitude.readthedocs.io/en/stable/rules/missing-accessibility-statement/

As noted, these pages make arguments related to clarity, but I disagree with these rules.

DanStoneMO (approved):

PR #286, opened 2026-06-05
Improve cmake targets for code linters

matthewrmshin:
Add target to run `fortitude check` if `fortitude` executable is found.

Modify cpplint target to use `cpplint` from the environment.

~Remove cpplint.py from lfric-jedi's source tree.~ (To be done separately.)

cpplint 2.0.2 happiness
- Update filter categories.
- Fix new violations, `build/include_what_you_use`.

Fix CI cmakelint logic and violations.

See also:
- https://github.com/JCSDA-internal/lfric-jedi/pull/1283
- https://github.com/JCSDA-internal/ops-um-jedi/pull/272
- MetOffice/mo-bundle#1098


.cmakelintrc
```diff
@@ -0,0 +1,2 @@
+filter=-readability/mixedcase
+linelength=133
```
mikecooke77, line 2: Any particular reason for a 133 line length?

mikecooke77 (approved):
It looks like the changes are all good at a glance but testing is the key.  Just one question which doesn't affect approval.

.cmakelintrc
matthewrmshin: The number is chosen to minimise changes.

DanStoneMO (approved):

PR #290, opened 2026-07-16
Update CI self hosted runner setting

matthewrmshin:
More fine grain group and labels.

See also:

- MetOffice/ops-um-jedi#283


ctgh (commented):
I don't understand the yaml syntax enough to comment on the change, so can you please explain roughly what it is doing?

matthewrmshin:
The syntax is recommended by a member of the Software Asset Management team to allow our action to pick up our runner using its runner group name as well as a label, so the path to chosing a runner should be less ambiguous.

ctgh (approved):
Thanks!

DJDavies2 (approved):

PR #293, opened 2026-08-11
Remove stats file production

### LINKED ISSUE #291: Failure opening stats file
I got this failure:

```console
<a href="http://www-nwp/~opsrc/OPS/view/dev/doc/gen_open_stats_file.html">gen_open_stats_file</a>
<span style="color: red">FATAL ERROR</span>
/home/users/david.davies/cylc-run/mo-bundle-415/log/job/1/ctest_opsinputs__spice_nag/01/job.stats: cannot open: I/O error on unit 7: No such file or directory
```

This is transient, but I suppose the question here is do we need stats files?  Can we just remove this functionality?  Looking at the code this isn't actually failing to write job.stats, it is failing to write a file called fort.7, which looks like it was a hack to cover up some other issue.

DJDavies2:
Fix #291.  It does a bit more than fix that issue in that it removes stats file generation altogether.  This doesn't affect the ctests as they continue to pass but if there is some use for stats file generation from opsinputs that isn't covered by this then this change isn't going to be appropriate.

To be clear, the problem in #291 is that because the stats file is written to fort.7, running opsinputs in parallel can therefore result in concurrent writes to the same file although this is a transient failure.

Test output:

https://cylchub/services/cylc-review/taskjobs/david.davies/?suite=OpsinputsStatsFile


matthewrmshin (approved):

adammaycock (approved):
I believe the "stats" file contains only a line informing how many obs were output to varobs. I can't see any use for this, so happy with this change.

deps/ops/public/GenMod_Control/GenMod_Control.F90
mikecooke77, line 1: I think now this is modified it should move to the stubs directory as it has diverged from the original file copied from OPS.

deps/ops/public/GenMod_Core/Gen_ReadConfigNL.inc
mikecooke77, line 1: I think now this is modified it should move to the stubs directory as it has diverged from the original file copied from OPS.

deps/ops/public/GenMod_Core/GenMod_Core.F90
mikecooke77, line 1: I think now this is modified it should move to the stubs directory as it has diverged from the original file copied from OPS.

mikecooke77 (changes requested):
I can see why these need to be removed.  Is this functionality that was used in OPS?  Could it have utility here?  

I have made one comment about the location of the changes files but apart from that I'm fine with this change.

DJDavies2:
> I can see why these need to be removed. Is this functionality that was used in OPS? Could it have utility here?
> 
> I have made one comment about the location of the changes files but apart from that I'm fine with this change.

Stats files were certainly used by people, they are produced operationally and were used to check on things like obs counts and obs stats.  @adammaycock probably knows more about these uses than me but they are/were certainly used.

However if we are working with ODB's I would assume that in a Jedi context there isn't anything in a stats file that cannot be obtained from an ODB.  This is probably true of OPS as well, but stats files predate ODB in OPS by many years.

The CODB metdb to odb tasks that generate odb's via OpsScr_CreateODB still generate stats files, although all they do is write out ob counts.  However that code is not affected as that is still using OPS code.

deps/ops/public/GenMod_Control/GenMod_Control.F90
DJDavies2: Do you mean just this file or the whole directory?

deps/ops/public/GenMod_Core/Gen_ReadConfigNL.inc
DJDavies2: Do you mean just this file or the whole directory?

deps/ops/public/GenMod_Core/GenMod_Core.F90
DJDavies2: Do you mean just this file or the whole directory?

adammaycock:
> > I can see why these need to be removed. Is this functionality that was used in OPS? Could it have utility here?
> > I have made one comment about the location of the changes files but apart from that I'm fine with this change.
> 
> Stats files were certainly used by people, they are produced operationally and were used to check on things like obs counts and obs stats. Adam Maycock (@adammaycock) probably knows more about these uses than me but they are/were certainly used.
> 
> However if we are working with ODB's I would assume that in a Jedi context there isn't anything in a stats file that cannot be obtained from an ODB. This is probably true of OPS as well, but stats files predate ODB in OPS by many years.
> 
> The CODB metdb to odb tasks that generate odb's via OpsScr_CreateODB still generate stats files, although all they do is write out ob counts. However that code is not affected as that is still using OPS code.

The stats files generated by OPS were certainly useful and used for a number of monitoring applications. They contained mean / RMS o-b stats, and rejection percentages etc. The stats files currently generated by JOPA don't contain any of that, they only contain a "Total number of Varobs" line which is of limited use. So I don't see a problem with getting rid, as proposed here.

For info, there is an open discussion on getting OPS like stats from JOPA: MetOffice/jedi-discussion#71.

matthewrmshin (approved):

mikecooke77:
I think when Wojciech created this the idea was for OPS files that are modified in opsinputs they should be in stubs.  For those files that are still comparable with OPS they should be kept where they are.  This was to help if and when we sync the OPS code with this repository.

DJDavies2:
Has such syncing ever happened?  I looked through the changelog for the past couple of years and didn't see anything, although it might not have been described as such in the summaries.  There were changes related to OPS, such as addition of support for EPSMWS, but these look like they have just been manually added to opsinputs, not synced in any way.

I think a policy of syncing something like OPS code in to opsinputs is dangerous.  Code in OPS runs in a particular context and environment and it is not tested outside of that environment and no consideration is given to it running outside of that environment by developers.  The stats file is a good example of this: there are top level Fortran programs that ensure that the file is create and closed as appropriate and that information is passed down to subroutines that write to it such as unit numbers.  On top of that there are scripts that control locations and names of stats file via environment variables.  It looks like opsinputs doesn't have any of this so it isn't surprising that this code is having difficulty.

mikecooke77:
> Has such syncing ever happened? I looked through the changelog for the past couple of years and didn't see anything, although it might not have been described as such in the summaries. There were changes related to OPS, such as addition of support for EPSMWS, but these look like they have just been manually added to opsinputs, not synced in any way.
> 
> I think a policy of syncing something like OPS code in to opsinputs is dangerous. Code in OPS runs in a particular context and environment and it is not tested outside of that environment and no consideration is given to it running outside of that environment by developers. The stats file is a good example of this: there are top level Fortran programs that ensure that the file is create and closed as appropriate and that information is passed down to subroutines that write to it such as unit numbers. On top of that there are scripts that control locations and names of stats file via environment variables. It looks like opsinputs doesn't have any of this so it isn't surprising that this code is having difficulty.

I don't think its been done recently (since Wojciech left) but there are scripts to do this type of syncing e.g. https://github.com/MetOffice/opsinputs/blob/develop/deps/update_ops_sources.sh.  In this one the code in "code" and "public" are syncing, the stubs are not.  Perhaps these scripts are no longer useful and potentially dangerous.

